PR #10, opened 2016-07-13
Add eslint and fix lint warnings

ithinkihaveacat:
Have globally disabled the require-jsdoc rule, and also (in two places) the requirement that functions with capital letters should only be used as constructors. (express wants to do this.)


andreban:
LGTM

PR #11, opened 2016-07-13
Update installation steps

andreban:
Improving installation steps on the docs to cover the installation and configuration of the Google Cloud Project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,30 @@
-# Guilliver
+# Gulliver
 
 A PWA Directory example.
+
+## Requirements
+1. [NodeJS](https://nodejs.org). 
+2. [Google Cloud SDK](https://cloud.google.com/sdk/)
+3. A project on the [Google Cloud Console](https://console.cloud.google.com/).
 
 ## Installation
 
 1. Fork the repository.
-2. Install [NodeJS](https://nodejs.org).
-3. Setup the repository.
+2. Clone the repository on your computer.
+3. Run `gcloud init` and follow the on-screen instructions. 
+4. Run `npm i` to install dependencies.
+5. Start the project with `npm start`
 
-```none
+Summary:
+
+```shell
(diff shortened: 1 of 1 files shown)


andreban:
@ithinkihaveacat PTAL

README.md
```diff
@@ -1,21 +1,30 @@
...
 A PWA Directory example.
 
+## Requirements
+1. [NodeJS](https://nodejs.org). 
+2. [Google Cloud SDK](https://cloud.google.com/sdk/)
+3. A project on the [Google Cloud Console](https://console.cloud.google.com/).
+
 ## Installation
 
 1. Fork the repository.
```
ithinkihaveacat: Is this needed? You can just clone, right?

README.md
```diff
@@ -1,21 +1,30 @@
...
+```shell
 $ git clone https://github.com/GoogleChrome/gulliver.git
 $ cd gulliver
+$ gcloud init
 $ npm i
 $ npm start
 ```
-
-For this to work, `gcloud` must be configured to point to a project you have the appropriate access to. Otherwise, the app will fail with a mysterious `Error: Not authorized.` error message. To check this, run `gcloud info`. If an unexpected account and project are listed, fix via `gcloud auth login` and/or `gcloud config set project [projectid]`.
+For this to work, `gcloud` must be configured to point to a project you have the appropriate access to. Otherwise, the app will fail with a mysterious `Error: Not authorized.` error message. To check this, run `gcloud info`. If an unexpected account and project are listed, fix via `gcloud init` and choose the `Re-initialize this configuration [default] with new settings` option.
```
ithinkihaveacat: Maybe either completely remove this (since the instructions now cover setting up gcloud) or move to a "Debugging" section and rewriting slightly?

ithinkihaveacat:
A few comments, LGTM though.

README.md
andreban: I think we can remove it.

PR #13, opened 2016-07-13
Add no-var rule to eslint

andreban:
Added the no-var rule to eslint and removed all
instances from the code.


andreban:
@ithinkihaveacat @juliantoledo PTAL

ithinkihaveacat:
LGTM

PR #14, opened 2016-07-14
Switch to Lighthouse's eslint configuration

ithinkihaveacat:
Too much pick-and-mix of coding styles? At least this way we're harmonized with one other GoogleChrome project, instead of none.


andreban:
LGTM

PR #16, opened 2016-07-14
Improve app structure

andreban:
Moving handlers to `/controllers`. This will make it easier to add
more handlers in the future and avoid breaking stuff in `app.js`.


ithinkihaveacat:
LGTM

PR #18, opened 2016-07-14
Add sudo: required

ithinkihaveacat:
Unfortunately for some reason we need both `dist: trusty` *and* `sudo: required` to get [the Trusty beta build environment](https://docs.travis-ci.com/user/trusty-ci-environment/)...


andreban:
LGTM

PR #19, opened 2016-07-15
Add simple client-side signin/signout

ithinkihaveacat:
Adds simple signin/signout feature. The buttons work, and actually sign the user in and out, but otherwise aren't connect to anything. (They don't disable any of the "edit" features, for example.)


public/js/gulliver.js
```diff
@@ -0,0 +1,95 @@
...
+      console.log('user signed out/never signed in');
+      login.disabled = false;
+      logout.disabled = true;
+    }
+  }
+
+  /* eslint-disable camelcase */
+  const params = {
+    scope: 'email',
+    client_id: '605287872172-js6omne47i1k79hnfo7d4bdu9rlemslr.apps.googleusercontent.com',
```
ithinkihaveacat: Where should we store the corresponding secret?

public/js/gulliver.js
```diff
@@ -0,0 +1,95 @@
+/* eslint-env browser */
```
ithinkihaveacat: Maybe don't need this quite yet, but need to use some sort of module system. (At the moment, everything is imported into global scope.)

andreban:
LGTM

views/base.jade
```diff
@@ -18,6 +18,14 @@ html(lang="en")
     meta(charset='utf-8')
     meta(name="viewport", content="width=device-width, initial-scale=1")
     link(rel="stylesheet", href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.2/css/bootstrap.min.css")
+    script.
+      window.gapiReady = new Promise(resolve => {
```
ithinkihaveacat: Kinda horrible but I think we need something like this if loading the `gapi` library async.

public/js/gulliver.js
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Note: window.gapiReady must be a promise that resolves to
+ * window.gapi.
+ *
+ * @template T
+ * @param {string} name the library to load
+ * @return {Promise<T>}
+ */
+function gapiLoad(name) {
+  return new Promise((resolve, reject) => {
```
ithinkihaveacat: ES6 Promises and arrow notation here, and elsewhere. For this code to run everywhere, we can either write ES5 by hand, or add a transpiler. If a transpiler (which I'd prefer), need to add this to the build process. (And also have a build process…)

public/js/gulliver.js
andreban: The `client_id` and `secret` look like something that should be stored in a config file. A sample of the config file can be saved in the repo and users will have to change and add their own key/secret to run the project. What are the best practices to this in Node? I'm guessing either a `JSON` file or a `config.js` that creates globals?

ithinkihaveacat: I don't think there's a standard way, just a bunch of options.

I guess the client_id should ultimately come from
https://github.com/GoogleChrome/gulliver/blob/master/config/config.json
though I don't know of a nice way to propagate it to the frontend.

Some possibilities:

1. Template the JS file itself, and run through templating system when
publishing. (Unexpected, and templating code is nasty.)
1. Load the config.json from the JS (synchonously). (Slow, maybe okay for
now.)
1. Make a templated/generated config.js (basically just returns
config.json), load via blocking `<script>`. (Maybe okay for now.)
1. Add a small chunk of config JS to the top of every HTML page. (Messy,
need to make inline JS executable.)
1. Add a module system, then do something similar to the NodeJS side
(`const config = require('config')`). (Needs module system. Probably the
"right" way, but needs a module system.)

ithinkihaveacat:
@andreban @juliantoledo Can you have a look at this? I think it's ready to go. Any further thoughts on where to store the secret?

public/js/gulliver.js
andreban: +1 to the build process + transpiler.

juliantoledo:
@ithinkihaveacat very cool, I will take a look in a bit!

public/js/gulliver.js
```diff
@@ -0,0 +1,95 @@
...
+  const [login, logout] = [document.getElementById('login'), document.getElementById('logout')];
+
+  /**
+   * @param {GoogleUser} user
+   */
+  function onChange(user) {
+    if (user.isSignedIn()) {
+      // @TODO Send id_token to backend
+      console.log('id_token', user.getAuthResponse().id_token);
+      login.disabled = true;
```
andreban: Would showing / hiding the login / logout buttons depending on the user state work better?

views/includes/header.hbs
```diff
@@ -15,10 +15,11 @@
 <div class="navbar navbar-default">
     <div class="container">
         <div class="navbar-header">
-            <div class="navbar-brand">Gulliver</div>
+            <div class="navbar-brand"><a href="/">Gulliver</a></div>
         </div>
         <ul class="nav navbar-nav">
-            <li><a href="/pwas">PWAs</a></li>
+            <li><button id="login" type="button" class="btn btn-default btn-sm" disabled>Login</button></li>
```
andreban: Not to be decided in this PR, but we need to figure out how the login flow fits in the UX. Maybe they will be slightly hidden since most users wont login. But it needs to make sense as part of the "Add a PWA flow".

andreban:
LGTM

juliantoledo:
LGTM

PR #20, opened 2016-07-15
Fix crash when PWA is added/created

ithinkihaveacat:
Creating a PWA crashes the server for me, I think because [crud.js#L87](https://github.com/GoogleChrome/gulliver/blob/master/controllers/pwas/crud.js#L87) doesn't actually pass a callback.


andreban:
LGTM

Issue #15, opened 2016-07-14
Remove unneeded dependencies


ithinkihaveacat:
There's a bunch of dependencies in `package.json` that we don't actually use/need. Remove them.

PR #21, opened 2016-07-15
Remove unused dependencies

ithinkihaveacat:
See #15.


ithinkihaveacat:
Still a bigger `node_modules` than I would like but still, one does one's best…

````sh
$ ls node_modules | wc -l
     342
$ du -h -d 0 node_modules
 91M	node_modules
````

andreban:
LGTM

PR #22, opened 2016-07-15
Request manifest via mobile UA

ithinkihaveacat:
PWAs such as Flipkart to UA detection, and will only serve a manifest to mobile UAs.

(Test URL: https://www.flipkart.com/manifest.json.)

### DIFF
--- a/controllers/pwas/crud.js
+++ b/controllers/pwas/crud.js
@@ -89,7 +89,17 @@ router.post(
         return next(err);
       }
 
-      fetch(savedData.manifestUrl).then(response => {
+      const options = {
+        method: 'GET',
+        headers: {
+          'user-agent': [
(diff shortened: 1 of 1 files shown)


paullewis:
FWIW Lighthouse can get you the manifest. We have a "save assets" option, and we can extend that to include the manifest, and get it back to you. We already request the page with a mobile UA.

ithinkihaveacat:
@paullewis Thanks yes I think ultimately (soon!) we want to avoid doing any of this ourselves, and deferring to Lighthouse. Am just making this quick fix so that actually trying to add `https://www.flipkart.com/manifest.json` doesn't lead to confusion, as it did for me.

(Thanks for looking over the repo BTW! Any/all drive-by comments appreciated!)

andreban:
LGTM

Issue #8, opened 2016-07-13
Switch templating format to handlebars


andreban:

ithinkihaveacat:
React might be an interesting option here too.

ithinkihaveacat:
Some pros and cons of different templating systems (following offline discussion with @andreban):

* Jade
  * Pro: already in place.
  * Con: unfamiliar (and disliked?) syntax? Not widely used.
* Mustache (Handlebars)
  * Pro: popular, familiar, simple.
  * Con: no databinding.
* Polymer
  * Pro: some usage by PWAs ([Konga](https://kongax.konga.com)).
  * Con: more than just templates, rarely rendered server-side (?).
* React
  * Pro: very popular, especially by people implementing PWAs ([Flipkart](https://www.flipkart.com), [Selio](https://www.selio.com), [NFL Now](https://www.nfl.com/now)).
  * Con: more than just templates, difficult to render server-side (?).

juliantoledo:
Mustache works very well for AMP pages, I think it fits what we need at the beginning. I only added Jade temporarily to get us started, but I prefer not to use it.

Let me read more about React to find out more about if it can handle server-side and how.

andreban:
I'm looking into handlebars. It seems that using it for including html sections into the page is not straightforward.

andreban:
jade has actually been rebranded as pug: https://github.com/pugjs/pug

andreban:
Fixed in #23

PR #23, opened 2016-07-18
Include Handlebars

andreban:
- Added handlebars and hbs to package.json.
- Added hbs templates for list, view and form.
- Add hbs as a template engine.
- Removed base jade template and templates for list, view and form.
- Removed jade as a template engine from app.js.
- Removed jade from package.json.


andreban:
This closes #8

adewale:
LGTM.
This seems to look exactly the same as the Jade version but I can't be sure.

Can you set up a staging server so people can see what the PRs look like?

PR #36, opened 2016-07-25
Use model in api.js

andreban:
- Changed the api.js module to re-use methods
from models/pwa.js

@ithinkihaveacat PTAL


ithinkihaveacat:
LGTM

PR #37, opened 2016-07-26
Standardize on arrow functions

andreban:
Switched to use arrow functions on callbacks.

PR #38, opened 2016-07-26
Use SVG for PWA image

andreban:
Using a SVG with background color + first letter of the name of the PWA as the default PWA image. In the future we'll use the PWA icon whenever possible.

### DIFF
--- a/app.js
+++ b/app.js
@@ -19,7 +19,7 @@ const path = require('path');
 const express = require('express');
 const config = require('./config/config');
 const hbs = require('hbs');
-
+const helpers = require('./views/helpers');
 const app = express();
 
 app.disable('etag');
@@ -27,6 +27,7 @@ app.set('views', path.join(__dirname, 'views'));
 app.set('view engine', 'hbs');
 app.set('trust proxy', true);
 hbs.registerPartials(path.join(__dirname, '/views/includes/'));
+helpers.registerHelpers(hbs);
 
 // Static files
 app.use(express.static('public'));
--- a/controllers/pwas/crud.js
+++ b/controllers/pwas/crud.js
@@ -31,6 +31,11 @@ router.get('/', (req, res, next) => {
     if (err) {
       return next(err);
     }
+
+    for (let pwa of entities) {
+      pwa.firstletter = pwa.name[0];
+    }
+
     res.render('pwas/list.hbs', {
       pwas: entities,
       nextPageToken: cursor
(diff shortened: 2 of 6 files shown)


andreban:
@ithinkihaveacat PTAL

app.js
```diff
@@ -38,12 +38,12 @@ app.use(require('./middlewares'));
 app.use(require('./controllers'));
 
 // Basic 404 handler
-app.use(function(req, res) {
+app.use((req, res) => {
```
ithinkihaveacat: Don't know if it's possible to break this out now, but this PR seems to have all changes from #37 in it as well.

views/pwas/list.hbs
```diff
@@ -24,11 +24,15 @@
             {{#each pwas}}
             <div class="media">
                 <a href="/pwas/{{id}}">
-                    <div class="media-left"><img src="/pwa.png">
+                    <div class="media-left">
+                        <svg width="64" height="64">
+                            <rect width="64" height="64" stroke-width="4" fill="{{backgroundColor}}" />
+                            <text x="32" y="32" fill="white" alignment-baseline="middle" text-anchor="middle" font-size="32px">{{firstletter}}</text>
```
ithinkihaveacat: Not completely sure about this, but it might be better to use a [helper](http://handlebarsjs.com/#helpers) for the `firstletter` stuff, so that this would turn into something like `{{firstLetter name}}`. Maybe depends on what other transformations we want to do.

controllers/pwas/crud.js
```diff
@@ -126,12 +131,13 @@ router.post(
...
  * Display a PWA.
  */
-router.get('/:pwa', function get(req, res, next) {
+router.get('/:pwa', (req, res, next) => {
   pwaModel.find(req.params.pwa, (err, entity) => {
     if (err) {
       return next(err);
     }
 
+    entity.firstletter = entity.name[0];
```
ithinkihaveacat: I think it would be better to either do these entity transformations and enhancements either in a separate function, or with a [helper](http://handlebarsjs.com/#helpers) at render time--it's not really clear why this is happening here.

app.js
andreban: It doesn. once the other one is merged, you will see only the correct ones here.

andreban:
@ithinkihaveacat Moved the functions to handlebar helpers. PTAL.

ithinkihaveacat:
LGTM!

PR #40, opened 2016-07-28
Use Icons from Manifest on Layout

andreban:
- Refactored manifest.js to be a class.
- Added Icons and background colors to UI.
- Need to fetch the icons and serve from Gulliver server

### DIFF
--- a/models/manifest.js
+++ b/models/manifest.js
@@ -16,26 +16,62 @@
 'use strict';
 
 const fetch = require('node-fetch');
+const DOMAIN_REGEXP = /(http[s]*:\/\/[a-z0-9A-Z-\.]+)(\/(.*?\/)*)*/;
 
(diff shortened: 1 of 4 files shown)


andreban:
@ithinkihaveacat PTAL

models/manifest.js
```diff
@@ -16,26 +16,100 @@
...
+    }
+
+    return domain + path + iconUrl;
+  }
+
+  static fetch(manifestUrl, callback) {
+    const options = {
+      method: 'GET',
+      headers: {
+        'user-agent': [
```
juliantoledo: let's move the user agent to a const at the top.

juliantoledo:
LGTM, Just one small comment.

andreban:
Moved the user-agent to a static variable. @juliantoledo  PTAL.

juliantoledo:
LGTM

PR #42, opened 2016-08-03
Serve images from GCloud

andreban:
- Fetches Icon and saves to a Bucked in GCloud
- Serves this Icon instead of directly from pwa server

### DIFF
--- a/lib/images.js
+++ b/lib/images.js
@@ -36,7 +36,7 @@ function fetchAndSave(url, destFile) {
         resolve(url);
       })
       .catch(err => {
-        reject.log(err);
+        reject(err);
       });
   });
 }
--- a/lib/images.js
+++ b/lib/images.js
@@ -26,6 +26,13 @@ const storage = gcloud.storage({
 const bucket = storage.bucket(CLOUD_BUCKET);
 const fetch = require('node-fetch');
 
+/**
+ * Fetches and Saves a Image to GCloud.
+ *
+ * @param url - a String with an URL to fetch the image.
+ * @param destFile - A string with the name of the Destination file on GCloud.
+ * @return a Promise.
+ */
 function fetchAndSave(url, destFile) {
   return new Promise((resolve, reject) => {
     fetch(url)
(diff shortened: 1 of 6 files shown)


andreban:
@juliantoledo PTAL

lib/images.js
```diff
@@ -24,62 +24,46 @@ const storage = gcloud.storage({
...
+  return new Promise((resolve, reject) => {
+    fetch(url)
+      .then(response => {
+        return saveImage(response, destFile);
+      })
+      .then(url => {
+        resolve(url);
+      })
+      .catch(err => {
+        reject.log(err);
```
ithinkihaveacat: Should this be `reject(err)`?

lib/images.js
```diff
@@ -24,62 +24,46 @@ const storage = gcloud.storage({
...
-
-  const gcsname = Date.now() + req.file.originalname;
-  const file = bucket.file(gcsname);
-  const stream = file.createWriteStream();
-
-  stream.on('error', err => {
-    req.file.cloudStorageError = err;
-    next(err);
-  });
+function saveImage(response, destFile) {
```
ithinkihaveacat: @juliantoledo @andreban WDYT about adding jsdoc to this and other functions? It would be useful to know what response is (a `http.IncomingMessage`?) and what `destFile` is (a string?), what the return value is, etc.

juliantoledo: Big +1 for jsdoc. Especially for anything in lib/

lib/images.js
andreban: Fixed

lib/images.js
andreban: Added JSDoc to `lib/images.js`. We should add to other parts soon.

juliantoledo:
I got a "T" temporarily when adding a new PWA. I guess we can later on refresh it on client side?

lib/images.js
```diff
@@ -24,62 +24,68 @@ const storage = gcloud.storage({
...
+const fetch = require('node-fetch');
 
-// Returns the public, anonymously accessable URL to a given Cloud Storage
-// object.
-// The object's ACL has to be set to public read.
-// [START public_url]
-function getPublicUrl(filename) {
-  return 'https://storage.googleapis.com/' + CLOUD_BUCKET + '/' + filename;
+/**
+ * Fetches and Saves a Image to GCloud.
```
juliantoledo: small nit, typo: s/a/an/

andreban: fixed

juliantoledo:
LGTM

PR #44, opened 2016-08-04
Add structure to write code test

andreban:
- Added mocha and istanbul to provide test and coverage reports
- Added initial tests for views/helpers

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ $ npm i
 $ npm start
 ```
 
+## Running Tests
+
+1. `./node_modules/.bin/mocha test/**/*.js` to run all tests.
+2. `./node_modules/.bin/istanbul cover _mocha test/**/*.js` for test coverage report.
+
 ## License
 
 See [LICENSE](./LICENSE) for more.
--- a/README.md
+++ b/README.md
@@ -26,8 +26,9 @@ $ npm start
 
 ## Running Tests
 
-1. `./node_modules/.bin/mocha test/**/*.js` to run all tests.
-2. `./node_modules/.bin/istanbul cover _mocha test/**/*.js` for test coverage report.
+1. `npm test` to run lint + tests + coverage report.
+2. `npm run mocha` to run all the tests only.
+3. `npm run coverage` to run tests + coverage report.
 
 ## License
 
--- a/models/pwa.js
+++ b/models/pwa.js
@@ -16,6 +16,7 @@
 'use strict';
 
 const config = require('../config/config');
+const images = require('../lib/images');
 const Manifest = require('./manifest');
 const db = require('../lib/model-' + config.get('DATA_BACKEND'));
 const gcloud = require('gcloud');
@@ -29,7 +30,6 @@ function mergeManifest(pwa, manifest) {
   pwa.startUrl = manifest.start_url || '';
   pwa.backgroundColor = manifest.background_color || '#ffffff';
   pwa.manifest = JSON.stringify(manifest);
-  pwa.iconUrl = manifest.getBestIconUrl();
 }
 
 exports.list = function(numResults, pageToken, callback) {
@@ -43,7 +43,7 @@ exports.find = function(key, callback) {
     }
 
     // Transform manifest into useful data.
-    data.manifestData = JSON.parse(data.manifest);
+    data.manifestData = Manifest.fromJson(data.url, JSON.parse(data.manifest));
 
     callback(null, data);
   });
@@ -78,7 +78,7 @@ exports.save = function(pwa, callback) {
       return callback(err);
     }
 
-    if (existingPwa && (!pwa.id || existingPwa.data.id.toString() !== pwa.id)) {
+    if (existingPwa && (!pwa.id || existingPwa.data.id.toString() !== pwa.id.toString())) {
       return callback(
           'Manifest already Registered for a different PWA', null);
     }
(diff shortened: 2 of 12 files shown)


andreban:
@juliantoledo @ithinkihaveacat What do you think about adding tests?

README.md
```diff
@@ -24,6 +24,11 @@ $ npm i
 $ npm start
 ```
 
+## Running Tests
+
+1. `./node_modules/.bin/mocha test/**/*.js` to run all tests.
```
ithinkihaveacat: The `mocha` script in `package.json` should also be modified to run this. (And likewise the `test` script should be changed to `npm run lint && npm run mocha`.)

README.md
```diff
@@ -24,6 +24,11 @@ $ npm i
...
+2. `./node_modules/.bin/istanbul cover _mocha test/**/*.js` for test coverage report.
```
ithinkihaveacat: I guess this should be a script as well. Maybe `coverage`? Not sure what's standard.

test/views/helpers/index.js
```diff
@@ -0,0 +1,36 @@
+var assert = require('assert');
+var helpers = require('../../../views/helpers');
+
+
+describe('views.helpers', function() {
```
ithinkihaveacat: Do we want arrow functions here too, or does it not matter?

andreban: Fixed

README.md
andreban: Updated package.json to run mocha and coverage.

andreban:
@ithinkihaveacat PTAL

juliantoledo:
LGTM

test/views/helpers/index.js
```diff
@@ -0,0 +1,36 @@
...
+describe('views.helpers', () => {
+  describe('#contrastColor', () => {
+    it('should return "white" when the value is not present', () => {
+      assert.equal('white', helpers.contrastColor(null));      
+      assert.equal('white', helpers.contrastColor(''));
+    });
+
+    it('should return "black" when value is "#FFFFFFF"', () => {      
+      assert.equal('black', helpers.contrastColor("FFFFFF"));      
+      assert.equal('black', helpers.contrastColor("#FFFFFF"));
```
sebastianbenz: what about #fff ?

sebastianbenz: What about upper lower case?

test/views/helpers/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+
+describe('views.helpers', () => {
+  describe('#contrastColor', () => {
+    it('should return "white" when the value is not present', () => {
+      assert.equal('white', helpers.contrastColor(null));      
+      assert.equal('white', helpers.contrastColor(''));
+    });
+
+    it('should return "black" when value is "#FFFFFFF"', () => {
```
sebastianbenz: test description does not matches what's tested -> description should be on a higher abstraction level than the code (otherwise it doesn't provide any value), better:

```
describe('#contrastColor', () => {
  describe('rgb color parsing', () => {
     it('supports '#' as prefix', => {});
     it('supports no prefix', => {});
  }
  it('generates contrast colors', () => {
     assert.equal('black', helpers.contrastColor("#FFFFFF"));
     assert.equal('white', helpers.contrastColor("#00000"));
  })
```

test/views/helpers/index.js
```diff
@@ -0,0 +1,36 @@
+var assert = require('assert');
+var helpers = require('../../../views/helpers');
+
+
+describe('views.helpers', () => {
+  describe('#contrastColor', () => {
+    it('should return "white" when the value is not present', () => {
```
sebastianbenz: nit: don't use 'should' it just adds noise, instead: 'returns "white" if value not present'

test/views/helpers/index.js
```diff
@@ -0,0 +1,36 @@
...
+  describe("#firstLetter", function() {
+    it('should return an empty string when value is not present', () => {
+      assert.equal('', helpers.firstLetter(null));
+    });
+
+    it('should return "G" for "Gulliver"', () => {
+      assert.equal('G', helpers.firstLetter("Gulliver"));
+    });
+
+    it('should return "G" for "gulliver"', () => {
```
sebastianbenz: test name does not match whats tested, see above

models/pwa.js
```diff
@@ -88,11 +88,37 @@ exports.save = function(pwa, callback) {
...
+        if (!err) {
+          updateIcon(savedPwa, manifest);
+        }
+        callback(err, savedPwa);
+      });
     });
   });
 };
 
+function updateIcon(pwa, manifest) {
```
sebastianbenz: why is there no test for this?

sebastianbenz:
added a few testing style nits - feel free to ignore

test/views/helpers/index.js
ithinkihaveacat: The `return` is gonna be duplicated even if the `should` is removed though, right? Still getting my head around this style of test…

sebastianbenz: In this case its not about duplication but about expressing the same shorter. My rule of thumb is: for a given `it "..."`, prepending all parent `describes` from the outer to the inner should result in a sentence. 

e.g. "views.helpers#contrastColor rgb color parsing supports '#' as prefix"

Issue #46, opened 2016-08-04
Set Content-Type on GCloud when uploading Images


andreban:
Since content-type is not being set, SVG images are being downloaded as application/octet-stream and failing to render. The content-type should be set as the same content type found on the response from the original server.

PR #48, opened 2016-08-05
Set Content-Type to file on GCloud

andreban:
- Uses the content-type from to response to set the content
  type on the GCloud file.

- The API available needs a second network call. I was looking for
  a way go set the Content-Type when creating the file or streaming
  it, but it seems this is the only way.


andreban:
Closes #46

juliantoledo:
LGTM

Issue #49, opened 2016-08-05
Fix manifest link on cards


juliantoledo:
Fix manifest link on cards, it currently shows in blue and does not use the calculated color like the Tittle uses. This makes it difficult to read in some backgrounds (strong yellow, brown, etc).

PR #50, opened 2016-08-05
Fix manifest link on cards #49

juliantoledo:
The links were difficult to read, I am inheriting the color to overwrite the link's blue.


andreban:
LGTM

PR #51, opened 2016-08-05
Set Content-Type when creating WriteStream

andreban:
- Found a way to set the Content-Type when creating the WriteStream.
  This avoids a second network call and will make uploading the
  image faster.


ithinkihaveacat:
LVGTM

PR #52, opened 2016-08-05
Fix eslint errors

ithinkihaveacat:
@andreban what there any reason to disable eslint for `/test`? I think we may as well have it…

### DIFF
--- a/test/views/helpers/index.js
+++ b/test/views/helpers/index.js
@@ -1,36 +1,37 @@
+/* global describe it */
+
 var assert = require('assert');
 var helpers = require('../../../views/helpers');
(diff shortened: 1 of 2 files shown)


test/views/helpers/index.js
```diff
@@ -1,36 +1,37 @@
+/* global describe it */
```
ithinkihaveacat: Probably some more "global" way to do this?

andreban:
I was having trouble getting `describe` and `it` to pass the eslint validation. Seems like `/* global describe it */` works. One more trick learned. LGTM!!

PR #53, opened 2016-08-05
Revert "Set Content-Type when creating WriteStream"

andreban:
Reverts GoogleChrome/gulliver#51

Doesn't actually work. My browser was caching the file.


andreban:
Unfortunately, doesn't actually work :(

PR #55, opened 2016-08-05
Verify id token

ithinkihaveacat:
Verify id token on the backend. Doesn't actually do anything with it yet (like associate it with a session).

We need some place to store the `CLIENT_SECRET`.

### DIFF
--- a/controllers/pwas/login.js
+++ b/controllers/pwas/login.js
@@ -48,7 +48,7 @@ function verifyIdToken(clientId, clientSecret, idToken) {
 router.use(bodyParser.json());
 
 // Handle middleware errors
-router.use((err, req, res, next) => { // eslint-disable-line no-unused-vars 
+router.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
   console.error(err.stack);
   res.status(400).send({
     error: 'couldn\'t parse JSON'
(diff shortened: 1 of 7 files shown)


controllers/pwas/login.js
```diff
@@ -0,0 +1,82 @@
+/**
```
andreban: I think this should live under `controllers/login.js` or `controllers/login/index.js` instead of `controllers/pwas/login.js`

controllers/pwas/login.js
```diff
@@ -0,0 +1,82 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// const CLIENTID = '605287872172-js6omne47i1k79hnfo7d4bdu9rlemslr.apps.googleusercontent.com';
+// const CLIENTSECRET = 'jc3R9JIRCes-yeEYqjnGKrhL';
+
+const express = require('express');
+const bodyParser = require('body-parser');
```
andreban: bodyParser is being required and used on [app.js](https://github.com/GoogleChrome/gulliver/blob/master/app.js#L33). Does it need to be included and used again here?

andreban: Ok. Answer is yes, since we're retrieving it as a JSON.

controllers/pwas/login.js
ithinkihaveacat: Moved.

andreban:
LGTM

PR #56, opened 2016-08-08
Add tests and use Promises for model-datastore.js

andreban:
- Using Promises for model-datastore.js
- Updated models/pwa.js to work with the Promises version
- Added tests for model-datastore.js

### DIFF
--- a/lib/model-datastore.js
+++ b/lib/model-datastore.js
@@ -22,50 +22,61 @@ const ds = gcloud.datastore({
   projectId: config.get('GCLOUD_PROJECT')
 });
 
-// Translates from Datastore's entity format to
-// the format expected by the application.
-//
-// Datastore format:
-//   {
-//     key: [kind, id],
-//     data: {
-//       property: value
-//     }
-//   }
-//
-// Application format:
-//   {
-//     id: id,
-//     property: value
-//   }
+/**
+ * Translates from Datastore's entity format to
+ * the format expected by the application.
+ *
+ * Datastore format:
+ *   {
+ *     key: [kind, id],
+ *     data: {
+ *       property: value
+ *     }
+ *   }
+ *
+ * Application format:
+ *   {
+ *     id: id,
+ *     property: value
+ *   }
+ *
+ * @param {Object} obj
+ * @return {Object}
+ */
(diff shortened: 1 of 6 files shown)


lib/model-datastore.js
```diff
@@ -82,75 +93,118 @@ function toDatastore(obj, nonIndexed) {
...
-      return cb(err);
-    }
-    const hasMore = entities.length === limit ? nextQuery.startVal : false;
-    cb(null, entities.map(fromDatastore), hasMore);
+    ds.runQuery(q, (err, entities, nextQuery) => {
+      if (err) {
+        return reject(err);
+      }
+      const hasMore = entities.length === limit ? nextQuery.startVal : false;
+      resolve(entities.map(fromDatastore), hasMore);
```
ithinkihaveacat: Does `hasMore` actually get returned here? I think you can only pass one argument to `resolve()`.

andreban: You are right. Should probably `resolve({entities: entities.map(fromDatastore), hasMore:hasMore};`

andreban:
CI is failing because I'm missing `CLIENT_SECRET` from `config.json`. Seems to be a side-effect of adding code that tests the datastore and reads the config. Should we add a placeholder for the `CLIENT_SECRET` on the `config.json`?

ithinkihaveacat:
@andreban We could disable the check but I think it would be better to add a dummy environment variable via `.travis.yml` ([docs](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-.travis.yml)):

```yaml
- env:
CLIENT_ID=dummy
CLIENT_SECRET=dummy
```

When we have tests that require non-dummy values we'll need to [generate encryption keys](https://docs.travis-ci.com/user/encryption-keys/).

andreban:
Fixed errors with environment variables, but the build is still breaking on the database tests. It seems to be related with network connections / IPv6:

`I0809 09:07:46.619410467    3797 socket_utils_common_posix.c:170] Disabling AF_INET6 sockets because ::1 is not available.`

ithinkihaveacat:
Test run fine for me locally.

Let's get this in and iterate?

LGTM

PR #59, opened 2016-08-11
Only allow logged in users to submit a PWA

andreban:
- Added `express-session` to support saving objects to session.
- User data is saved to the session, after `/api/login` is called.
- `crud.js` checks for the `session.user` availability before
   adding PWA
- `pwa.js` checks for the existence of `user.id` and `user.email`
   before persisting to database..


andreban:
@ithinkihaveacat @juliantoledo 

Added this to start the discussion. There are a few things that need to be decided
- We should probably use some sort of session persistence, since express-session's MemorySession is not supposed to be used in production. Not sure if gcloud has a plugin for this. Will take a look. Ideas are welcome. 
- I haven't touched editing a PWA. Right now I'm wondering if it makes sense, since the only data that is submitted is the Manifest URL. Maybe this should be more of an Admin thing?
- Same thing for deleting a PWA. We probably don't want anyone deleting a PWA. This also looks like and admin thing to me. (Or at least only allowing the submitter + admins to delete it).
- We need to add code to the submission page that enables / disables the "Submit" button, given the user state. Another option is catching the error in the backend and redirecting to the submission page again, with an error message.

andreban:
@ithinkihaveacat @juliantoledo I think this is more the way we want it now.

controllers/pwas/crud.js
```diff
@@ -58,37 +61,78 @@ router.get('/add', (req, res) => {
...
+  verifyIdToken(CLIENT_ID, CLIENT_SECRET, idToken)
+    .then(user => {
+      const pwa = {
+        manifestUrl: manifestUrl,
+        createdOn: new Date(),
+        updatedOn: new Date(),
+        user: {
+          id: user.getPayload().sub,
+          email: user.getPayload().email
+        }
```
ithinkihaveacat: @juliantoledo Do we want to store the email? I think in the absence of a good use for it, we should avoid saving it. Might be also a good idea to HMAC the id as well, especially if it's not obvious that a submission is linked to a user.

juliantoledo: @andreban @ithinkihaveacat Yes, I think is a good practice to do both.

juliantoledo:
LGTM (with the comment about email and sha1 of the id).

andreban:
Removed e-mail and using SHA-1 to key the userId. @ithinkihaveacat PTAL.

juliantoledo:
LGTM, I tested the user flow and the sha1 in the db. We can keep your names createdOn and updatedOn when we merge (I used updated, created in the other PR).

PR #60, opened 2016-08-16
Improve getting started experience

ithinkihaveacat:
Still not perfect, but this should eliminate some categories of surprising error message.


juliantoledo:
LGTM

Issue #69, opened 2016-08-19
Eliminate the offline dinosaur


ithinkihaveacat:
Every page is dynamically generated, so it's going to be difficult to make anything meaningful happen offline. However, the offline dinosaur should never appear.

juliantoledo:
We should also make sure we show an icon/message when offline.

ithinkihaveacat:
Now that #131 is merged, I think this is done.

PR #76, opened 2016-08-24
Add simple service worker and sw-toolbox

ithinkihaveacat:
Part of #69.

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -15,7 +15,8 @@
 
 'use strict';
 
-const router = require('express').Router(); // eslint-disable-line
+const express = require('express');
+const router = express.Router(); // eslint-disable-line
 
 // PWAs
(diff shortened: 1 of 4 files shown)


andreban:
Are we considering using `sw-precache` for the static content? If so, we might want to add a build process (eg: gulp)?

ithinkihaveacat:
@andreban I did look at `sw-precache` but I think `sw-toolbox` gives slightly better control. And I was hoping that for now we could hand-write the configuration. (Might change my mind after doing a bit more work though.)

And I think we will need a build tool at some point, but hope to get away without it for now.

juliantoledo:
+1 one on gulp, but also +1 on doing it later. I will take a deeper look tomorrow.

juliantoledo:
LGTM

PR #77, opened 2016-08-25
Make 404 handler prettier

ithinkihaveacat:
Tidy up the 404 handler. (Need something similar for #69.)

### DIFF
--- a/controllers/pwas/crud.js
+++ b/controllers/pwas/crud.js
@@ -136,7 +136,8 @@ router.post('/:pwa/edit', (req, res, next) => {
 router.get('/:pwa', (req, res, next) => {
   pwaModel.find(req.params.pwa, (err, entity) => {
     if (err) {
-      return next(err);
+      // Not really an error: the pwa wasn't found in the db. Fall through to 404 page.
+      return next();
     }
 
     res.render('pwas/view.hbs', {
--- a/app.js
+++ b/app.js
@@ -41,14 +41,13 @@ app.use(require('./middlewares'));
 // Controllers
 app.use(require('./controllers'));
 
-// Basic 404 handler
+// If no route has matched, return 404
 app.use((req, res) => {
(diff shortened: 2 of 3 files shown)


controllers/pwas/crud.js
```diff
@@ -136,7 +136,8 @@ router.post('/:pwa/edit', (req, res, next) => {
 router.get('/:pwa', (req, res, next) => {
   pwaModel.find(req.params.pwa, (err, entity) => {
     if (err) {
-      return next(err);
+      // Not really an error: the pwa wasn't found in the db. Fall through to 404 page.
```
andreban: Should the find method invoke the callback(null, null), so both err and entity are null? Maybe it's more clear in the code than testing an error that's not exactly an error.

andreban:
LGTM, with a comment that we can deal with later.

juliantoledo:
LGTM.

Let's also create another PR with tests for all the entry points, including this 404s.

controllers/pwas/crud.js
ithinkihaveacat: Strangely this error comes all the way from https://github.com/GoogleChrome/gulliver/blob/master/lib/model-datastore.js#L175. It appears `get()` does not distinguish between a "Bad Request" and a key not found.

Not sure of the "right" way to fix this--pre-flight the `get()` with a query for the key? But that would be two trips. @juliantoledo do you know?

Anyway, I don't think this is a problem for this PR, will punt for later.

juliantoledo: We should handle them differently I think, let's do a PR for error handling fixes.

ithinkihaveacat: @juliantoledo See #81.

ithinkihaveacat:
@juliantoledo See #80 for the tests. You want to add a milestone?

PR #82, opened 2016-08-26
Improve manifest parsing robustness

ithinkihaveacat:
We feel over parsing https://www.slideshare.net/manifest.json, because it doesn't have a `start_url`. This PR fixes this, and also saves the manifest itself for a time when we actually test this stuff.

/cc @adewale

### DIFF
--- a/models/pwa.js
+++ b/models/pwa.js
@@ -31,8 +31,8 @@ const E_MANIFEST_ERROR = exports.E_MANIFEST_ERROR = 2;
 
 function mergeManifest(pwa, manifest) {
   pwa.name = manifest.name;
-  pwa.startUrl = manifest.start_url || '';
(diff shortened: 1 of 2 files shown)


andreban:
This assumes that if the PWA hasn't provided a `start_url`, the `start_url` is it's homepage. In the context of linking to the PWA, I think the homepage is fine. But in the context of displaying what the `start_url` is, I'm not really happy with it. 

Maybe something like this:
```
  pwa.startUrl = manifest.start_url || ''; // This is what we display
  pwa.pwaLink = uri(manifest.start_url || '/').absoluteTo(manifest.url).toString(); //This is what we link to
```

andreban:
Btw, let's merge this and fix the crash first, then change anything if needed.

PR #86, opened 2016-08-31
Deep copy object before saving

andreban:
It looks like node-gcloud modifies Objects in some scenarios when
saving them to the datastore. The case I was able to reproduce is
it modifying arrays to be of type 'arrayValue' when the array is
inside another class. This case is represented in the included
test case.

On our current state, a shallow copy of the object would be enough,
but since the problem could arise again if we have the class
instance with an array deeper in the object structure. A test case
has also been added to cover this case.

### DIFF
--- a/lib/model-datastore.js
+++ b/lib/model-datastore.js
@@ -106,6 +106,10 @@ function deepCopy(object) {
     return object;
   }
 
+  if (object instanceof Date) {
+    return object;
+  }
+
   const clone = Object.assign({}, object);
   Object.keys(clone).forEach(k => {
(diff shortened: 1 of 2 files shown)


andreban:
@juliantoledo @ithinkihaveacat PTAL.

juliantoledo:
Can you fix the lint error?
Do I need to delete my old db objects to test it? (otherwise I get an error).

andreban:
Fixed the lint error. You shouldn't have to delete any db object. It looks like the list test is flacky. Does the update one work for you?

juliantoledo:
LGTM

PR #87, opened 2016-09-01
Prevent offline dinosaur

ithinkihaveacat:
If offline, either show cached content or a special offline page (`/.shell/offline`).

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -2,40 +2,29 @@
 
 importScripts('/sw-toolbox/sw-toolbox.js'); /* global toolbox */
 
-self.addEventListener('install', () => {
-  console.log('INSTALL');
-});
+// The offline page itself *and* its dependencies
+const OFFLINE = [
+  '/.shell/offline',
+  '/sw.js',
+  '/js/gulliver.js'
+];
 
-self.addEventListener('activate', () => {
-  console.log('ACTIVATE');
-});
+const OFFLINE_URL = '/.shell/offline';
 
-self.addEventListener('fetch', e => {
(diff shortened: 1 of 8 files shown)


public/sw.js
```diff
@@ -1,15 +1,30 @@
 /* eslint-env serviceworker, browser */
 
-importScripts('/sw-toolbox/sw-toolbox.js');
+importScripts('/sw-toolbox/sw-toolbox.js'); /* global toolbox */
 
-self.addEventListener('install', () => {
-  console.log('INSTALL');
-});
+// The offline page itself *and* its dependencies
```
ithinkihaveacat: This is annoying. Don't see any easy way to generate this. (Annoying that `/.shell/offline` needs to be a real endpoint, too.)

juliantoledo:
@ithinkihaveacat Nice progress with offline!
I think we should disable the login buttons.

Force reload shows the offline dinosaur, if I reload again it shows the page. Is that normal for sw?

I also found that if the DB is offline the page hangs on the server, we should do something about that in another PR, maybe let the sw kick in after a timeout.

andreban:
I agree with disabling login buttons.

@juliantoledo force-reload showing the offline dinosaur is expected. It makes the browser skip the service-worker. Regular reload should work.

It's possible to tweak timeouts in 'sw-toolbox'. I think it's important in order to solve the LieFi scenario. But agreed we can file an issue and do it on a following PR.

juliantoledo:
LGTM, it's working really good now.

I will open a feature request to grey out PWAs that are not available offline.

PR #94, opened 2016-09-02
Use meta / manifest description

andreban:
Changed description getter to give preference to the manifest
description and use the metaDescription when the first is not
available


juliantoledo:
LGTM

PR #96, opened 2016-09-02
Redirect HTTP requests to HTTPS

### LINKED ISSUE #85: Make sure HTTP redirect to HTTPS
I found that in the testing app engine domain the redirection is not automatic (unlike the development internal version).

andreban:
- Updated app.yaml

Fixes #85


juliantoledo:
LGTM

We can investigate later to find out why Lighthouse does not read it properly as HTTPS, I don't know where the problem is.

Issue #98, opened 2016-09-02
Fix pwa edit page showing [Object] in Manifest field


andreban:
Manifest content is showing as Object. Replace `pwa.manifest` with `pwa.manifestAsString`.

<img width="379" alt="screen shot 2016-09-02 at 4 34 19 pm" src="https://cloud.githubusercontent.com/assets/1733592/18209329/3b591162-712b-11e6-9d06-852223bb3cc9.png">

PR #101, opened 2016-09-02
Show manifest correctly on edit form

andreban:
The code was using `pwa.manifest` that is now an object. Switched to
using `pwa.manifestAsString` instead.

Fixe #98


juliantoledo:
LGTM

PR #102, opened 2016-09-02
Set metaDescription correctly in lib/pwa.js

### LINKED ISSUE #100: Set 'metaDescription' correctly in pwa.js
The code is trying to set pwa.description, which produces the following error:

`[TypeError: Cannot set property description of #<Pwa> which has only a getter]`

Switch to use set `pwa.metaDescription` instead.

andreban:
Fixes #100


juliantoledo:
LGTM

Issue #26, opened 2016-07-18
Integrate with Lighthouse


andreban:
Integrate with Lighthouse to:
- [x] Get Lighthouse score

PR #103, opened 2016-09-02
Lighthouse integration

juliantoledo:
Lighthouse integration #26 steps:
- [x] Lighthouse model class
- [x] Lighthouse json data fetcher  
- [x] Move Lighthouse server URL to config.json
- [x] Store Lighthouse info
- [x] Generate Lighthouse info when new PWA is added

I will leave for another PR:
- Entry point to fire tasks: POST /task/lighthouse/task/:pwaid
- Batch processing
- Scheduling


lib/data-fetcher.js
```diff
@@ -18,6 +18,8 @@
 var fetch = require('node-fetch');
 const cheerio = require('cheerio');
 
+const LIGHTHOUSE_SERVER = 'https://api-dot-pwa-directory.appspot.com/audit?url=URL&fmt=json';
```
andreban: Maybe we could move this to `config.json`.

juliantoledo: Yes, will do.

andreban:
@adewale FYI

juliantoledo:
It's failing one of the two test setups, the one with "node_js": "4"

 Uncaught TypeError: path must be a string
      at TypeError (native)
      at Object.fs.readFile (fs.js:250:11)
      at test/lib/lighthouse.js:29:12

@andreban @ithinkihaveacat any ides on how the path should be? It works with nodejs 6.

controllers/task.js
```diff
@@ -0,0 +1,57 @@
...
+const express = require('express');
+const lighthouseLib = require('../lib/lighthouse');
+const router = express.Router(); // eslint-disable-line new-cap
+
+/**
+ * POST /task/lighthouse/:pwaid
+ *
+ * Fetch and store a lighthouse report for the pwaid
+ */
+router.post('/lighthouse/:pwaid', (req, res, next) => {
```
andreban: Is there any way we can only allow AppEngine to start tasks?

juliantoledo: Yes, I think so, I will check tomorrow.

juliantoledo: @andreban I think the two options seems to be either OAuth2 or ReverseDns the source IP.

andreban: Reverse DNS looks easy: https://nodejs.org/api/dns.html#dns_dns_reverse_ip_callback

andreban:
LGTM with the comment on checking who's invoking the task when not running from `localhost`.

juliantoledo:
@andreban great, I will look into the reverseDNS and tasks queues.

juliantoledo:
@andreban We may be able to run the batches without the task endpoint and I am also not sure if reverseDNS will work. Removed and I am leaving that for a another PR.

PR #104, opened 2016-09-02
Move client_id into configuration

ithinkihaveacat:
Slightly cleaner, I think--we gain a global but avoid having to embed client ids into JS. There will probably be other values we need to make accessible to the front-end too.


andreban:
LGTM

PR #110, opened 2016-09-05
Add visible create / updated dates

andreban:
- Add created and updated fields on Pwa constructor.
- Update the updated field when a Pwa is updated.

### DIFF
--- a/views/pwas/view.hbs
+++ b/views/pwas/view.hbs
@@ -48,8 +48,8 @@
         <div class="pwabox-body">
           <h3>{{pwa.name}}</h4>
           <p>{{pwa.description}}</p>
-          <p>Added: {{pwa.created}}</p>
-          <p>Updated: {{pwa.updated}}</p>          
+          <p>Added {{moment pwa.created}}</p>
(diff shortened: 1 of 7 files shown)


views/pwas/view.hbs
```diff
@@ -48,6 +48,8 @@
         <div class="pwabox-body">
           <h3>{{pwa.name}}</h4>
           <p>{{pwa.description}}</p>
+          <p>Added: {{pwa.created}}</p>
```
andreban: How should we format the dates? Is just year, month and day enough or we also want to add hour + minute?

ithinkihaveacat: moment.js [does relative dates](http://momentjs.com/docs/#/displaying/fromnow/), if you want to escalate to that…

andreban: That looks like a good one. "Created 5 days ago" definitely reads better.

andreban:
Added `moments.js` to format the dates. Looks awesome now. Thanks @ithinkihaveacat

ithinkihaveacat:
LGTM

Issue #72, opened 2016-08-22
Add Pagination support


juliantoledo:
We need to add pagination support

andreban:
I was looking at pagination using gcloud. It seems that finding out the number of pages [may be tricky](http://www.shiftedup.com/2015/03/08/how-to-count-all-entries-of-a-given-type-in-the-app-engine-datastore) using cloud-datastore. I'm not really happy with any solutions given in the article.

I've fixed the existing "More" button. We may need to add a "Previous Page".

andreban:
I think we should migrate pagination to use skip / limit and if the database grows enough that it starts being an issue, we change? WDYT? @juliantoledo @ithinkihaveacat

andreban:
Changed to skip / limit on #160

andreban:
Closed by #160, improved by #161

andreban:
Closed on #160 and #161. UI update being tracked on #141

PR #112, opened 2016-09-05
Fix Pagination code

andreban:
Fixes the existing pagination.

Related to #72

### DIFF
--- a/lib/model-datastore.js
+++ b/lib/model-datastore.js
@@ -139,7 +139,8 @@ function list(kind, limit, token) {
       if (err) {
         return reject(err);
       }
-      const hasMore = entities.length === limit ? nextQuery.startVal : false;
+      const hasMore = nextQuery.moreResults === 'MORE_RESULTS_AFTER_LIMIT' ?
+          nextQuery.endCursor : false;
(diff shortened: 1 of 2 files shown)


juliantoledo:
How can I test it?

andreban:
@juliantoledo add more than 10 PWA's to your instance or change the `LIST_PAGE_SIZE`on `crud.js` to a small number (I've been testing with `1`).

juliantoledo:
@andreban Yes, I was able to test it now. How do you go back? should we have another link?

andreban:
Check my comment on #72 

The preferred way to handle pagination in cloud-datastore is using the cursor pointer, since using offset/limit scales linearly. 

The issue is that I don't get a pointer to the previous page, only the next. We could store this in the session or something but it would still break if someone sends the link to other person.

juliantoledo:
We should take a deeper look before adding this then.

juliantoledo:
@andreban would it be possible to keep the pointer to previews page as URL parameter?

andreban:
Closing this as we're going back to the skip/limit solution.

Issue #71, opened 2016-08-22
Various sort options: name, newest, score, etc.


juliantoledo:
We should add various sort options: by name, newest, score, etc.

PR #113, opened 2016-09-05
Adds backend support for sorting

andreban:
- ?sort=newest and ?sort=name

Starts work on #71


ithinkihaveacat:
LGTM, although how can this be tested? http://localhost:8080/?sort=newest gives "No PWAs found." Do we need to re-initialize the database or something?

juliantoledo:
I added one and I saw it in newest. Should we make newest the latest X PWAs instead?

andreban:
@ithinkihaveacat - the `created` field was added on a recent PR. The PWAs on your DB probably don't have this field. The interesting part is that when you call `order` by a field and the record doesn't have it, `cloud-datastore` doesn't return the record.

PR #117, opened 2016-09-05
Precache first page visited

### LINKED ISSUE #116: Save first page loaded to offline cache
The very first page the user sees is not saved to the offline cache. That is, if you visit URL1, then click on URL2, and then go offline, URL1 is not saved to the offline cache.

This is a consequence of the service worker lifecycle: the very first page load is not "controlled" (i.e. processed) by the service worker, so it never enters the cache.

ithinkihaveacat:
Fixes #116, which can be reproduced via:

1. Clear cache, service workers, etc.
2. Load http://localhost:8080/.
3. Click on a PWA to load a different URL.
4. Go offline.
5. Hit "back" in the browser.

Expected behavior:

http://localhost:8080 appears as in step (2).

Actual behavior:

(Custom) offline page appears.

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -2,16 +2,26 @@
 
 importScripts('/sw-toolbox/sw-toolbox.js'); /* global toolbox */
 
-// The offline page itself *and* its dependencies
+// URL to return in place of the "offline dino" when client is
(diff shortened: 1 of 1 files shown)


juliantoledo:
LGTM

PR #118, opened 2016-09-07
Remove Edit and Delete endpoints

### LINKED ISSUE #114: Remove ability to edit/delete PWAs
screen shot 2016-09-05 at 5 37 43 pm

- [ ] From the web interface (for all users)
- [ ] From the API

andreban:
- Remove buttons from PWA view
- Remove put, delete and update from API
- Remove /:pwa/delete and /:pwa/edit

Closes #114

### DIFF
--- a/controllers/pwas/api.js
+++ b/controllers/pwas/api.js
@@ -43,21 +43,6 @@ router.get('/', (req, res, next) => {
 });
 
 /**
- * POST /api/pwas
- *
- * Create a new PWA.
- */
-router.post('/', (req, res, next) => {
-  pwaModel.save(req.body)
-    .then(entity => {
-      res.json(entity);
-    })
-    .catch(err => {
-      return next(err);
-    });
-});
-
-/**
  * GET /api/pwas/:id
  *
  * Retrieve a PWA.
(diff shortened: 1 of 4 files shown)


config/config.json
```diff
@@ -1,7 +1,7 @@
 {
-  "GCLOUD_PROJECT": "google.com:gulliver",
```
ithinkihaveacat: Did you mean to change these?

I override these via environment variables, that are set by [direnv](http://direnv.net/). (So you have an `.envrc` file that lists the environment variables I want sent, and then hook into the shell so that whenever I change into the gulliver directory they are sent, and whenever I change into some other directory they are unset. You can also source the "parent" `.envrc` to set `GIT_AUTHOR_EMAIL`, etc. Works quite well.)

An alternative might be to use git's [assume aliases](http://durdn.com/blog/2012/11/22/must-have-git-aliases-advanced-examples/#assume-aliases) to mark files you don't want to commit.

andreban: Didn't mean to change it. Reverted to the correct version. PTAL

ithinkihaveacat:
Aside from config change, LGTM.

PR #120, opened 2016-09-07
Add Manifest to view.hbs

andreban:
Result looks like this:

screen shot 2016-09-07 at 1 55 15 pm


juliantoledo:
@andreban I actually think we should show it in some kind of lightbox, maybe using pure CSS and having a raw link to the actual manifest.

juliantoledo:
What do you think about the lightbox or even make it css expandable? Do you prefer to do it in another PR?

andreban:
I think we can go with expanding? I'd rather do in another PR.

juliantoledo:
LGTM

Issue #115, opened 2016-09-05
Remove /api endpoint and associated code


ithinkihaveacat:

andreban:
I removed the endpoints that allow to add, edit or delete a PWA. Should we keep view and list?

PR #123, opened 2016-09-08
Remove /api endpoint and associated code #115

juliantoledo:
Removed the old api test we do not need at the moment and rename the crud.js controller to pwa.js


andreban:
Didn't run it, but code looks LGTM

PR #127, opened 2016-09-12
Fixing jsdoc of lib/data-fetcher and lib/pwa

juliantoledo:
Fixing jsdoc of lib/data-fetcher and lib/pwa


juliantoledo:
I will keep working on this one after the other PR is merged, because I added more methods to both classes.

juliantoledo:
I moved this to #128

Issue #129, opened 2016-09-13
UX consultation


ithinkihaveacat:
Among other things:

- [x] Detail page (#140)
- [x] Pagination (#141)
- [ ] Offline state (#142)
- [x] Icon (#125)

ithinkihaveacat:
(@andreban has ideas on detail page, pagination.)

juliantoledo:
Let's leave only #142 open.

PR #131, opened 2016-09-13
Show online/offline status in footer

ithinkihaveacat:
This addresses https://github.com/GoogleChrome/gulliver/issues/69#issuecomment-241420992.

Looks like this:

screen shot 2016-09-13 at 3 30 21 pm
screen shot 2016-09-13 at 3 30 16 pm

(Small CSS transition between states. Tried for a little while to get the offline status to slide up/down but wasn't working … will wait for UX input #129 before trying harder.)

PTAL @juliantoledo.

### DIFF
--- a/public/js/gulliver.js
+++ b/public/js/gulliver.js
@@ -186,6 +186,21 @@ function setupOnlineAware() {
       }
     });
   }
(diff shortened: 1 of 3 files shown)


juliantoledo:
LGTM

PR #160, opened 2016-09-19
Changes pagination to use skip / limit

andreban:


andreban:
Deployed with page-size = 1 here: https://gulliver-1371.appspot.com/

juliantoledo:
LGTM with this fixes:
@andreban I think we should use a page size like 12 (looks good with 1, 2, 3 and 4 columns) and is not too big for mobile (1 column). I also found that when the links is in the bottom of the page I cannot click it.

We can add some CSS to the links later in another PR as part of the footer that we should implement.

andreban:
@juliantoledo Updated with 12 PWA's. I haven't touched the frontend for the PWA listing. Just using whatever we had before. I think we can discuss the layout and do it in another PR.

PR #161, opened 2016-09-20
Add memcached to Gulliver

andreban:
- Adds cache when listing pwas
- Adds cache when retrieving a single PWA


juliantoledo:
From chat:
- [ ] Skip memcache if not avaiable
- [ ] Add memcache local installation to readme

andreban:
- Tweaked timeout to fail faster. This will print a message to console letting the developer know that memcached is not installed.
- Added instructions to Readme.
- Failure already worked, but was taking too long due to long timeouts / retries.

andreban:
PTAL

juliantoledo:
LGTM

PR #162, opened 2016-09-20
Lighthouse fixes

juliantoledo:
We are storing more information from the Lighthouse report to be used in the details page.


lib/lighthouse.js
```diff
@@ -81,28 +110,42 @@ exports.fetchAndSave = function(pwaId) {
  */
 exports.processLighthouseJson = function(lighthouseJson) {
   var lighthouseInfo = {};
-  lighthouseJson.aggregations.forEach(section => {
+  lighthouseJson.aggregations.forEach(agregation => {
```
andreban: s/agregation/aggregation

lib/lighthouse.js
```diff
@@ -81,28 +110,42 @@ exports.fetchAndSave = function(pwaId) {
  */
 exports.processLighthouseJson = function(lighthouseJson) {
```
andreban: This method has many instances of the usage of 'var'. We should always use 'const' / 'let'

lib/lighthouse.js
juliantoledo: done.

lib/lighthouse.js
juliantoledo: @andreban I will leave this change for #166.

andreban:
LGTM

Issue #149, opened 2016-09-15
Add a Web App Manifest validator


juliantoledo:
We should add a Web App Manifest validator.

For example: https://web.telegram.org/manifest.json is not valid because it does not have an array of icons.

andreban:
Their icons section looks like this:
```
  "icons": {
    "16": "img/icons/icon16.png",
    "32": "img/icons/icon32.png",
    "60": "img/icons/icon60.png",
    "64": "img/icons/icon64.png",
    "90": "img/icons/icon90.png",
    "128": "img/icons/icon128.png"
  }
```

I think we're expecting something like the following:
```
  "icons": [{
    "src": "/icon/web_hi_res_512.png",
    "sizes": "512x512",
    "type": "image/png"
  }]
```

Is their approach valid? If it is we should support. Otherwise, we can test if `icons` is an Array before using it.

juliantoledo:
Seems valid:
https://developer.chrome.com/extensions/manifest/icons

I wonder if we have a good way to validate the manifest.json formatting.

ithinkihaveacat:
We could probably use https://github.com/mounirlamouri/manifest-validator
(written by a Googler).

juliantoledo:
Seems like the telegram one is for the Chrome Web Store and not at Web App Manifest.

The icons need to be an array:
https://w3c.github.io/manifest/#icons-member

Let's try that one @ithinkihaveacat

brendankenny:
Lighthouse has a manifest parser that we forked from the one from mlamouri and have slowly updated for more corner cases/correctness and added tests: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/lib/manifest-parser.js

The format it spits out is a little confusing, because rather than log any errors to console, at every step into the parsed manifest object it gives the parsed `value` but also an error string (`debugString`). In practice you end up writing a lot of things like manifest.value.icons.value[0].value.src.value, but I can write a quick converter that would just spit out what the parsed value ends up being :)

brendankenny:
as an example, for your bad icons example, 

```js
  {"icons": {
    "16": "img/icons/icon16.png",
    "32": "img/icons/icon32.png",
    "60": "img/icons/icon60.png",
    "64": "img/icons/icon64.png",
    "90": "img/icons/icon90.png",
    "128": "img/icons/icon128.png"
  }}
```
gives a parsed response of (cutting out other manifest fields):
```js
{
    "icons": {
      "value": [],
[...]

PR #165, opened 2016-09-20
Validate manifest

ithinkihaveacat:
- Add Manifest validator from Lighthouse
- Minimal tests
- Integrate into front-end

(See #149.)

Note that the two JS files in `third_party` are imported from different GitHub repositories. (The `install.sh` script can be used to update them if necessary.)


lib/pwa.js
```diff
@@ -122,11 +122,12 @@ exports.findByManifestUrl = function(manifestUrl) {
  *
  * Steps:
```
ithinkihaveacat: This is super horrible. We should separate out the parts. (The fetching and the validation, at least.)

third_party/README.md
```diff
@@ -0,0 +1,2 @@
+See `./install.sh` for information on where the `*.js` files in this directory
```
ithinkihaveacat: `third_party/` contains imported code. We could also depend on lighthouse itself, but doesn't seem quite worth it. (And would still need to patch `manifest-parser.js`, I think.

juliantoledo:
Great, it validates https://web.telegram.org/manifest.json properly with Error: 'icons' expected to be an array but is not.

third_party/README.md
ithinkihaveacat: No, the necessary JS files are included in the repository. ./install.sh
only needs to be run to update to newer versions. (Was this comment
deleted? I couldn't find it on GH.)

juliantoledo:
LGTM

PR #166, opened 2016-09-20
Eliminate all var (change to let)

ithinkihaveacat:
Also add `eslint` rule to keep it that way!

(See https://github.com/GoogleChrome/gulliver/pull/162#pullrequestreview-730118.)


juliantoledo:
LGTM

Issue #137, opened 2016-09-13
Improve PWA Detail layout on Mobile


andreban:
Some of the content is being rendered outside the viewport on mobile.

<img width="386" alt="screen shot 2016-09-13 at 4 28 19 pm" src="https://cloud.githubusercontent.com/assets/1733592/18480040/1e97b98c-79cf-11e6-9af4-d1da6ebe4953.png">

juliantoledo:
This is part of  #140, I will do some quick changes before we add Tabs for Manifest and Lighthouse.

PR #169, opened 2016-09-20
Created a simple details page, while we build one with tabs.

juliantoledo:
Created a simple details page, while we build one with tabs for #137


andreban:
LGTM

Issue #133, opened 2016-09-13
Add Google Analytics


andreban:
Add GA to all pages. 

**Production**

```
<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-82966500-2', 'auto');
  ga('send', 'pageview');
</script>
```

**Development**

```
<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-82966500-1', 'auto');
  ga('send', 'pageview');

</script>
```

juliantoledo:
Development:
`<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-82966500-1', 'auto');
  ga('send', 'pageview');

</script>`

Production:
`<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-82966500-2', 'auto');
  ga('send', 'pageview');

</script>`

[...]

PR #172, opened 2016-09-21
Added Google Analytics for #133

juliantoledo:
Added Google Analytics to all pages and config.json:
"GOOGLE_ANALYTICS": "your Google Analytics tracking code"

### DIFF
--- a/controllers/pwa.js
+++ b/controllers/pwa.js
@@ -103,29 +103,17 @@ router.post('/add', (req, res, next) => {
       if (typeof err === 'number') {
         switch (err) {
           case pwaLib.E_MANIFEST_INVALID_URL:
-            res.render('pwas/form.hbs', {
-              pwa,
-              error: `pwa.manifestUrl [${pwa.manifestUrl}] is not a valid URL`
-            });
-            return;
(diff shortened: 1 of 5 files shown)


controllers/pwa.js
```diff
@@ -44,7 +45,8 @@ router.get('/', (req, res, next) => {
         hasPreviousPage: pageNumber > 1,
         nextPageNumber: pageNumber + 1,
         previousPageNumber: pageNumber - 1,
-        currentPageNumber: pageNumber
+        currentPageNumber: pageNumber,
+        GOOGLE_ANALYTICS: GOOGLE_ANALYTICS
```
ithinkihaveacat: I think if you add `GOOGLE_ANALYTICS` to [`app.locals`](https://github.com/GoogleChrome/gulliver/blob/cc77640bfd577464c583eb8871cc885ca5ed8d11/app.js#L35) it will be automatically available to all templates. (Then it wouldn't need to be added manually to every template.)

Might also be better to to give `GOOGLE_ANALYTICS` a camel case name, to match other variables.

juliantoledo: @ithinkihaveacat great tip. Done.

PR #173, opened 2016-09-22
Adds cache invalidation when a PWA is added

andreban:
- When a PWA is added / updated, the current  timestamp is added
to the Cache with the key `PWA_LIST_LAST_UPDATE`. This value
has the timeout of 5 times the Pwa List timeout.
- When a pwa list page is added to the Cache, a property
`cacheTimestamp` with the current timestamp is appended to the
object.
- When a page is loaded from the cache, the `PWA_LIST_LAST_UPDATE`
is also loaded.
- If no value is found for that page, the Promise
fails and the value is loaded from the db.
- If a value is found
from the key, but not found for `PWA_LIST_LAST_UPDATE`, the
timestamp has expired and the Pwa can be used, since the expire
time for the timestam is larger than the one for the PWA.
- If a value for the timestamp is found and is larger then the
PWA List cacheTimestamp, it means the value is expired and the
Promise fails.
- If a value for the timestamp is found and is smaller then the
PWA List cacheTimestamp, the PWA is valid and the Promise resolves.

### DIFF
--- a/lib/data-cache.js
+++ b/lib/data-cache.js
@@ -62,7 +62,20 @@ function set(key, value, timeout) {
   });
 }
 
+function getMulti(keys) {
+  return new Promise((resolve, reject) => {
+    memcached.getMulti(keys, (err, data) => {
+      if (err) {
+        console.error(err);
+        return reject(err);
+      }
+      return resolve(data);
+    });
+  });
+}
+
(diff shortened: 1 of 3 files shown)


juliantoledo:
LGTM

Issue #132, opened 2016-09-13
Create Tabs for Recently Added / Best Scores


andreban:
We already have the sorting in the backend. We can use Recently Added as the default tab and Top Scored as the second tab.

Decision:
  - Recently Added (sorted by creation date; paginate through to show all)
  - Top Scores (sorted by score; paginate through to show all)

juliantoledo:
Nice, we could have: All : Recent : Best Scores

PR #174, opened 2016-09-22
Tabs for Newest and Best Score 

juliantoledo:
The tabs get smaller with @media at 380px and 480px page width


juliantoledo:
Fixes #132

PR #178, opened 2016-09-22
Added Metadata to html

juliantoledo:


controllers/pwa.js
```diff
@@ -49,7 +53,9 @@ router.get('/', (req, res, next) => {
         sortOrder: sortOrder,
         showNewest: sortOrder !== 'newest',
         showScore: sortOrder !== 'score'
-      });
+      };
+      libMetadata.add(arg, req);
```
ithinkihaveacat: I'm not sure, but it might be cleaner if this is flipped around, so that the default arguments are retrieved first, and are then overwritten with specifics. So it would be something like:

```javascript
let arg = Object.assign(libMetadata.fromRequest(req), {
  title: 'PWA Directory',
  description: '...'
});
res.render('pwas/list.hbs', arg);
```

This would also eliminate the `addMissingMetadata()` method from `metadata.js`. It also wouldn't be necessary to pass `arg` to `libMetadata.add()`.

views/includes/metadata.hbs
```diff
@@ -0,0 +1,56 @@
...
+<meta property="og:description" content="{{description}}" />
+<meta property="og:site_name" content="PWA Directory" />
+<meta property="article:published_time" content="{{datePublished}}" />
+<meta property="article:modified_time" content="{{dateModified}}" />
+
+<!-- Metadata required by Google -->
+<script type="application/ld+json">
+{
+  "@context": "http://schema.org",
+  "@type": "NewsArticle",
```
ithinkihaveacat: `NewsArticle`?

views/includes/metadata.hbs
```diff
@@ -0,0 +1,56 @@
...
+                 "@type": "Person",
+                 "name": "Michael Stillwell"
+            },
+           {
+                 "@type": "Person",
+                 "name": "Andre Bandarra"
+            }],
+  "publisher": {
+    "@type": "Organization",
+    "name": "Google",
```
ithinkihaveacat: Should we be saying this? /cc @adewale

views/includes/metadata.hbs
juliantoledo: @ithinkihaveacat I don't know what other category we should use. Any ideas?

controllers/pwa.js
juliantoledo: @ithinkihaveacat good idea, thanks and fixed.

views/includes/metadata.hbs
juliantoledo: Changed to Google Developer Relations

views/includes/metadata.hbs
juliantoledo: I will change it to CollectionPage, we can decide later what is better if we want to have rich results.

PR #179, opened 2016-09-23
Style details page

amedina:
Changes to populate and style the details page:
- Add template partials
- Inject data from lighthouse report
- Add handlebars helpers
- Define CSS rules

### DIFF
--- a/views/helpers/index.js
+++ b/views/helpers/index.js
@@ -72,10 +72,29 @@ exports.highlightedJson = function(object) {
   return syntaxHighlight(JSON.stringify(object, null, 4));
 };
 
+exports.getAggregationTableRow = function(aggregation) {
+  var row = '<tr>'
+          + '<td>' + aggregation.name + '</td>'
+          + '<td>' + aggregation.overall + '</td>'
+          + '</tr>';
+  return row;
+};
+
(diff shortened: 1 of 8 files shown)


views/helpers/index.js
```diff
@@ -44,13 +44,60 @@ exports.moment = function(date) {
   return moment(date).fromNow();
 };
 
+/*
+  Light-weight helper function to generate
+  highlighted JSON strings
+*/
+function syntaxHighlight(json) {
```
ithinkihaveacat: We need to think about whether we should actually do this at all (see https://github.com/GoogleChrome/gulliver/issues/180).

views/includes/css.hbs
```diff
@@ -1,4 +1,4 @@
-{{!-- Copyright 2015-2016, Google, Inc.
+/* Copyright 2015-2016, Google, Inc.
```
ithinkihaveacat: What was the reason for changing this? (Don't think it matter much either way, just curious.)

amedina: Two reasons: the file contents are CSS; and (2) I used autoprefixer to ensure coverage of all browsers and the HBS comment causes the tool to choke. Cheers.

Issue #180, opened 2016-09-23
Security review: user-generated content


ithinkihaveacat:
We display UGC on Gulliver (the manifest itself and lighthouse results) so we need to make sure you can't do anything bad with this.

We should use a security-minded library for escaping, and probably also add content-security-policy headers to prevent inline JS from being executed. (And we actually have some inline JS at the moment, so this will need to be removed.)

This will probably also affect #179, which does syntax highlighting. (We might not be able to do this.)

juliantoledo:
I will probably usa https://github.com/cure53/DOMPurify

For #179, Would moving the syntax highlighting to the server side help?

PR #184, opened 2016-09-26
Added sanitize to html fetching and CSP directives for #180

juliantoledo:


juliantoledo:
I added the sanitizer at the HTML fetcher for the Metadata description, is it needed anywhere else?

Apparently we do not need it for the manifest fetching because response.json does JSON.parse(), is that enough?

lib/data-fetcher.js
```diff
@@ -33,7 +43,7 @@ const USER_AGENT = ['Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) Appl
 function fetchMetadataDescription(url) {
   return new Promise((resolve, reject) => {
```
ithinkihaveacat: From a previous PR, but I think the `new Promise()` can be eliminated here. (It could written `return fetchWithUA(url)…`, with `return $('meta[name="description"]').attr('content');` in the final `.then()`, rather than resolve.)

lib/data-fetcher.js
```diff
@@ -33,7 +43,7 @@ const USER_AGENT = ['Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) Appl
...
     fetchWithUA(url)
-      .then(response => response.text())
+      .then(response => DOMPurify.sanitize(response.text()))
```
ithinkihaveacat: Do we need to protect `cheerio` from bogus input? I don't think it's going to do any fetching, etc. itself. Maybe sanitize the description only?

juliantoledo: done.

juliantoledo: @ithinkihaveacat I actually think is better to make sure cheerio gets a clean html just in case.

lib/data-fetcher.js
juliantoledo: done.

PR #188, opened 2016-09-27
404 page fixes and 500 errors fixes

juliantoledo:


app.js
```diff
@@ -62,14 +62,14 @@ app.use(require('./controllers'));
 
 // If no route has matched, return 404
 app.use((req, res) => {
-  res.status(404).render('404.hbs');
+  res.status(404).render('404.hbs', {nonce1: req.nonce1, nonce2: req.nonce2});
 });
 
 // Basic error handler
-app.use((err, req, res) => {
+app.use((err, req, res, _) => {
```
juliantoledo: I need the 4th variable to get this code executed, I was using next but that gave lint errors, it works with underscore.

ithinkihaveacat: Ugh, gotcha. Forgot about that magic express feature. Maybe add a comment to explain?

juliantoledo: @ithinkihaveacat done. thanks!

views/pwas/view.hbs
```diff
@@ -19,7 +19,7 @@
   <body>
     {{> header}}    
     <div class="section-title">
-      <a href="javascript:history.back()"><div class="back"><i class="fa fa-chevron-left" aria-hidden="true"></i></div></a>
+      <a href="{{referer}}"><div class="back"><i class="fa fa-chevron-left" aria-hidden="true"></i></div></a>
```
andreban: If an external site links to this page in Gulliver, when the user hits back, she will return to the referring website. I'm just wondering if this is the desired behaviour or if the desired behaviour would be to navigate to the parent page (The PWA Listing).

juliantoledo: Good catch, I could filter it and send them back to our home page, but I am fine sending them back to the referer as well.

juliantoledo:
@andreban added a fix to send user to home if referer is null or from a different host.

views/pwas/view.hbs
juliantoledo: fixed.

controllers/pwa.js
```diff
@@ -159,6 +159,11 @@ router.post('/add', (req, res, next) => {
  * Display a PWA.
  */
 router.get('/:pwa', (req, res, next) => {
+  const host = req.get('host');
+  const homeUrl = req.protocol + '://' + host;
+  let referer = req.headers.referer &&
```
ithinkihaveacat: Maybe `up`/`upurl`/`uplink` would be more semantic. Also, I think this can be `const`.

ithinkihaveacat: (Actually, probably not worth changing.)

andreban:
LGTM

PR #192, opened 2016-09-28
Make scheme explicit in CSP policy

ithinkihaveacat:
Safari 9 needs the scheme. (Via manual testing and https://scotthelme.co.uk/safari-doesnt-like-csp/.)


juliantoledo:
@ithinkihaveacat does *.domain work?

ithinkihaveacat:
@juliantoledo Good point, I don't know. Safari isn't complaining, but I realize now that's probably because it's not actually executing the GA code that would be affected (because of the nonce issue). Will test again after fixing #193.

PR #195, opened 2016-09-28
Generate gulliver.js (ES5) from gulliver.es6.js (ES6)

ithinkihaveacat:
Safari 9 can't handle the arrow functions we had in `gulliver.js`. Add a "build" step to generate an ES5 version.

One (unfortunate) implication of this means that when editing `gulliver.es6.js` we need to run `npm prestart` before reloading. (We should improve this later.)

Note that `npm start` runs this automatically, so `npm start` (and deploying to GCP?) should work just fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ npm-debug.log
 .jshintrc
 .idea/
 key.json
+public/js/gulliver.js
+public/js/gulliver.js.map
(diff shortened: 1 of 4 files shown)


juliantoledo:
I added /public/js/gulliver.js to .eslintignore
because the generated file uses tabs instead of spaces.

PR #197, opened 2016-09-29
Make Safari work

ithinkihaveacat:
- Makes scheme explicit
- Removes need for nonce


ithinkihaveacat:
(Need to rebase and fix merge conflicts.)

juliantoledo:
@ithinkihaveacat LGTM, good solution for the JS inline :)

juliantoledo:
@ithinkihaveacat is the 'unsafe-eval' still needed? I removed it locally and I don't see the error anymore, maybe the bug fix is deployed already.

PR #200, opened 2016-09-29
Fix offline state edge case

ithinkihaveacat:
The toggling of offline state for the PWAs did work on the very first page load.

To reproduce:

1. Clear caches, etc.
1. Load http://localhost:8080.
2. Disable network.

Actual:

PWAs are not greyed out.

Expected:

PWAs are grey out. (Since none visited yet.)

### DIFF
--- a/public/js/gulliver.es6.js
+++ b/public/js/gulliver.es6.js
@@ -136,12 +136,12 @@ function setupOnlineAware() {
       if (href) {
         fetch(href, {method: 'HEAD'}).then(r => {
           if (r.status === 200) {
-            // Offline, but available in cache
+            // Available in cache, allow click
             this.style.transition = 'opacity .5s ease-in-out';
             this.style.opacity = 1;
             this.onclick = null;
           } else {
-            // Offline, but not cached
+            // Not cached, prevent click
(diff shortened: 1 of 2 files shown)


juliantoledo:
@ithinkihaveacat the Best Score tab stays enabled and does not work, I get a blank page (with errors in the console), when navigating back all PWAs are enabled again until I reload the page.

Do we need to cache both?
http://localhost:8080/?sort=newest
http://localhost:8080/?sort=score

ithinkihaveacat:
We can cache both but I think it's more consistent to grey them out. So I did this, but then realized that there are even more links that should be unavailable:

screen shot 2016-10-26 at 12 56 41 pm

(The links in the header, and the "next page" link should also be greyed out--nothing is actually clickable if you visit the homepage, and then go offline.)

So I can fix all these, but it ends up being a pretty crappy/weird experience--in this case of no content being available it's probably better to put up a separate blocking modal dialog as soon as the user goes offline. (However, if some pages have been visited, and are available in the cache, it looks better to simply grey out the links that are unavailable.)

juliantoledo:
LGTM

PR #211, opened 2016-10-06
Optimize images serving

### LINKED ISSUE #145: Optimize PWA icon serving
For example:

- generate multiple icon sizes. The sizes being used in Gulliver right now are:
  - 128x128
  - 64x64
- optimize after fetch from origin

andreban:
- Use the 'sharp' library to create images with 64 and 128 sizes
- Updated tests

Resolves #145

### DIFF
--- a/lib/images.js
+++ b/lib/images.js
@@ -17,6 +17,7 @@
 
 const gcloud = require('google-cloud');
 const config = require('../config/config');
+const sharp = require('sharp');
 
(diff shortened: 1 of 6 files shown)


andreban:
I created the new icons sizes. Haven't implemented falling back to the original icon if the optimized is not available (for PWA's registered before this commit). Do you think it's worth doing it or should we just update all the PWAs?

juliantoledo:
@andreban LGTM, but let me work on the update process so we can merge this one.

Issue #121, opened 2016-09-07
Update Lighthouse info daily/weekly


juliantoledo:
We need to queue Lighthouse requests to update each PWAs score daily/weekly.

For tasks we need to test:
- [ ] bee-queue or kue
- [ ] Google Tasks Queues [https://github.com/google/google-api-nodejs-client/blob/master/apis/taskqueue/v1beta2.js](url)
- [ ] Cron/Scheduling

addyosmani:
If you run into any trouble getting batch Lighthouse processing for PWAs setup, feel free to reach out. @brendankenny has been exploring a local version of this and I started exploring a super rough version of this a week ago. Note that some of Lighthouse's tests when run headless may not be 100% accurate atm (e.g Service Worker/offline tests don't always score the way they do when run from the extension or command line).

brendankenny:
yes, I'm not sure about the exact status these days, but at least recently Service Workers were barely functional at all in headless Chrome

juliantoledo:
@addyosmani @brendankenny, We have been working with @crhym3, he created a docker image that runs Lighthouse with headless Chrome. We found few differences in the score/results, we may need help to improve that in the long term.

brendankenny:
Ah, yes, we had a thread a while back with @crhym3 over issues with headless he was running into. I think fundamentally it's still very much a work in progress and it's still a little hard to predict what won't be working and how that'll manifest when you run it :)

The good news is (I believe) Alex has said he can switch his setup over to using real Chrome if need be, so if the headless bugs pile up too deep we can always go with that.

x1ddos:
I'm updating server-side lighthouse version + headless today, hoping things will get better.
Heaving issue with re-compiling headless_shell though, seems to be broken.

PR #213, opened 2016-10-07
New optimized images and update tasks

juliantoledo:
In this change, we add two cron jobs tasks:
/tasks/cron creates one task for each PWA for its update once a week
/tasks/execute executes one of these tasks every minute

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -1,0 +1,80 @@
+/**
+ * Copyright 2015-2016, Google, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const express = require('express');
+const pwaLib = require('../lib/pwa');
+const tasksLib = require('../lib/tasks');
+const Task = require('../models/task');
+const router = express.Router(); // eslint-disable-line new-cap
+
+/**
+ * GET /task/cron
+ *
+ * We use a GET for the cron job to launch PWA update process.
+ */
+router.get('/cron', (req, res, next) => {
+  pwaLib.list()
+    .then(result => {
+      console.log();
+      for (let i = 0; i < result.pwas.length; i++) {
+        tasksLib.push(new Task(result.pwas[i].id));
+      }
+      res.sendStatus(200);
+    })
+    .catch(err => {
+      next(err);
+    });
+});
+
+/**
+ * GET /task/execute
+ *
+ * We use a GET for the cron job to launch PWA update process.
+ */
+router.get('/execute', (req, res, next) => {
+  tasksLib.pop()
+    .then(taks => {
+      if (taks) {
+        pwaLib.find(taks.pwaId)
+          .then(pwa => {
+            pwaLib.save(pwa);
+          })
+          .catch(err => {
+            next(err);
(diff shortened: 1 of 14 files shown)


juliantoledo:
Fixes #121

juliantoledo:

andreban:
This deprecates #211, right?

controllers/tasks.js
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * GET /tasks/cron
+ *
+ * We use a GET from the cron job to launch a PWA update process
+ * for all PWAs.
+ */
+router.get('/cron', (req, res, next) => {
+  // Only requests from App Engine Cron are allowed
+  if (req.get(APP_ENGINE_CRON)) {
```
andreban: Can we change the comment to: "Checks for the presence of the 'X-Appengine-Cron' header on the request. Only requests from App Engine are allowed"

controllers/tasks.js
```diff
@@ -0,0 +1,94 @@
...
+    pwaLib.list()
+      .then(result => {
+        // Create one taks for each PWA
+        for (let i = 0; i < result.pwas.length; i++) {
```
andreban: Could also be written as:
```
result.pwas.forEach(pwa => {
    tasksLib.push(new Task(pwa.id));
});
```

controllers/tasks.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * GET /tasks/execute
+ *
+ * We use a GET from the cron job to execute each PWA update task
+ */
+router.get('/execute', (req, res, next) => {
+  // Only requests from App Engine Cron are allowed
+  if (req.get(APP_ENGINE_CRON)) {
+    tasksLib.pop()
+      .then(taks => {
```
andreban: s/taks/task

lib/tasks.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+'use strict';
+
+const config = require('../config/config');
+const db = require('../lib/model-' + config.get('DATA_BACKEND'));
+const Task = require('../models/task');
+
+const ENTITY_NAME = 'Task';
+const E_SAVING_TAKS = exports.E_SAVING_TAKS = 1;
```
andreban: s/TAKS/TASK

controllers/tasks.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * GET /tasks/cron
+ *
+ * We use a GET from the cron job to launch a PWA update process
+ * for all PWAs.
+ */
+router.get('/cron', (req, res, next) => {
+  // Only requests from App Engine Cron are allowed
+  if (req.get(APP_ENGINE_CRON)) {
+    pwaLib.list()
```
andreban: This lists every PWA. We could write this like "Give me all PWA's that haven't been updated in the last 24 hours". This way, we could, for instance, have the cron job running every hour and every app would be updated each 24hs. It would also make it easier to change the behaviour (eg: update every app that hasn't been updated in the last 7 days. Run every hour).

controllers/tasks.js
juliantoledo: hum, I had it similar to this and lint complained but now it works.

controllers/tasks.js
juliantoledo: done

lib/tasks.js
juliantoledo: done

controllers/tasks.js
juliantoledo: Good idea, done.

controllers/tasks.js
andreban: We can probably put this as an Issue and implement later?

juliantoledo:
@andreban Fixed the comments, let's do another PR as you suggested for the daily task changes.

lib/tasks.js
```diff
@@ -21,8 +21,8 @@ const Task = require('../models/task');
 
 const ENTITY_NAME = 'Task';
 const E_SAVING_TAKS = exports.E_SAVING_TAKS = 1;
```
andreban: Missed this one: s/TAKS/TASK

PR #218, opened 2016-10-13
Fixes #217, Manifest without icons fails to run Lighthouse

### LINKED ISSUE #217: Manifest without icons fails to run Lighthouse
Manifest without icons fails to run Lighthouse, for example:
https://housing.com/manifest.webmanifest

There is an unhandled error:
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: Path must be a string. Received null

juliantoledo:


lib/pwa.js
```diff
@@ -278,12 +278,10 @@ exports._save = function(pwa) {
     })
     .then(savedPwa => {
       this.addPwaToCache(savedPwa);
-      this.updateIcon(savedPwa, pwa.manifest)
+      this.updateIcon(savedPwa, pwa.manifest);
+      this.updateLighthouseInfo(savedPwa)
```
andreban: The reason that updateLighthouseInfo is called on the .then part of updateIcon is to guarantee the order in which the steps are executed. Is this change needed to make it work? Or the code below alone should work?

juliantoledo: @andreban done

Issue #89, opened 2016-09-02
Add a build system to the project


andreban:
Add a build system such as `gulp` to the project. This will help avoiding things like:
- Adding express routes to serve js modules from node_modules, since the build can copy the files to /public
- Make it possible to use `sw-precache` for the static content

PR #219, opened 2016-10-13
Improve JS build process

ithinkihaveacat:
- Split out the gapi-related functions into a separate file
- Use rollup to remove unused code
- Babel instead of TypeScript to transpile JS
- Minify JS via uglify

Related to #89.


juliantoledo:
LGTM

PR #220, opened 2016-10-14
Make rollup and babel production deps

ithinkihaveacat:
*Whoops!* This PR makes `npm start` work if packages are installed via `npm install --production`. (It's fine with `npm install`, but that's probably not what's happening on GAE.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     ]
   },
   "dependencies": {
+    "babel-preset-es2015-rollup": "^1.2.0",
     "body-parser": "^1.15.2",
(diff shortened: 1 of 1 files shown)


ithinkihaveacat:
Would be nice to have an automated test for this. It's basically something like:

1. `npm install --production` — install modules
2. `npm start` — start the server

Is anything running `http://localhost:8080`?

Not sure how to express this in the Travis config, though. (It installs packages via `npm install`.)

juliantoledo:
LGTM

Issue #225, opened 2016-10-21
Indicate if no lighthouse score is available


sebastianbenz:
Before we have a first lighthouse score we show the score as 0. This can be misleading. Either show `-`, `n/a` or maybe an icon, e.g. [this](https://design.google.com/icons/#ic_hourglass_full).

juliantoledo:
Great idea. Thank you @sebastianbenz

ithinkihaveacat:
I suppose a push notification would be nice, once the score is available
too.

juliantoledo:
It takes about 1 minute to get the score, let's open an issue to start building notifications in general.

PR #228, opened 2016-10-24
Indicate that the Lighthouse is in progress #225

juliantoledo:


juliantoledo:

models/pwa.js
```diff
@@ -22,7 +22,7 @@ class Pwa {
   constructor(manifestUrl, manifestModel) {
     this.manifestUrl = manifestUrl;
     this.manifest = manifestModel;
-    this.lighthouseScore = 0;
+    this.lighthouseScore = undefined;
```
andreban: We can probably remove this line, since this.lighthouseScore will be undefined by default.

PR #230, opened 2016-10-25
Lighthouse graph

juliantoledo:


lib/lighthouse.js
```diff
@@ -72,6 +72,58 @@ exports.findByPwaId = function(pwaId) {
...
+
+/**
+ * Retrieves the Lighthouse Grpah data for a PWA
+ * in Google Charts JSON format
+ *
+ * @param {number} pwaId
+ * @return {Promise<Json>}
+ */
+exports.getLighthouseGraphByPwaId = function(pwaId) {
+  return new Promise((resolve, reject) => {
```
ithinkihaveacat: I think it should be possible to eliminate the `Promise` constructor and just `return this.getLighthouseByPwaId(…)`.

controllers/api.js
```diff
@@ -0,0 +1,47 @@
...
+    .then(lighthouseGraph => {
+      if (lighthouseGraph) {
+        res.setHeader('Cache-Control', 'public, max-age=' + CACHE_CONTROL_EXPIRES);
+        res.json(lighthouseGraph);
+      } else {
+        res.status(404);
+        res.json('not found');
+      }
+    })
+    .catch(err => {
```
ithinkihaveacat: We probably should tidy up this error handling in this and other controllers in another PR, see https://expressjs.com/en/advanced/best-practice-performance.html#use-promises.

.eslintignore
```diff
@@ -1,3 +1,4 @@
 /coverage
 /third_party
 /public/js/gulliver.js
+/public/js/ligthouse-chart.js
```
ithinkihaveacat: Also, why ignore this file?

juliantoledo: I certainly have a big problem spelling lighthouse :(

The problem is that 'google' is not defined  no-undef, because is imported in the page and not in this file, how can I solve that?

ithinkihaveacat: Oh, I see. I think the right thing to do is tell eslint `google` is a global via a comment. Actually, also needs the `browser` environment to get away with `XMLHttpRequest`:

```js
/* global google */
/* eslint-env browser */
```

public/js/ligthouse-chart.js
```diff
@@ -0,0 +1,58 @@
...
+      });
+    };
+    xhr.send();
+  });
+}
+
+function drawChart() {
+  const chartDiv = document.getElementById(CHART_ELEMENT_ID);
+  const pwaId = chartDiv.getAttribute("pwa");
+  console.log(pwaId);
```
juliantoledo: done.

public/js/ligthouse-chart.js
```diff
@@ -0,0 +1,58 @@
+const CHART_ELEMENT_ID = 'chart';
```
juliantoledo: done.

.eslintignore
juliantoledo: Thanks, done.

views/includes/lighthouse.hbs
```diff
@@ -11,6 +11,7 @@
     </div>
   </div>
   {{#if pwa.lighthouseScore}}
+  <div id='chart' pwa={{pwa.id}}></div>
```
juliantoledo: done.

public/js/ligthouse-chart.js
```diff
@@ -0,0 +1,58 @@
...
+const CHART_BASE_URL = '/api/lighthouse-graph/';
+
+function makeRequest(method, url) {
```
juliantoledo: done.

views/includes/head.hbs
```diff
@@ -30,10 +30,15 @@
...
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.min.css">
 
+<!-- Origin Trial Token, feature = Web Share, origin = https://pwa-directory.appspot.com, expires = 2016-12-06 -->
+<meta http-equiv="origin-trial" data-feature="Web Share" data-expires="2016-12-06" content="AuNW4MmMk4aT1N9HPDpSlUPkBywNReGpXpbTC+9nOE/rv6W6mToe7J1ZSmz35g+6BbPzjmE9gpBTZOrb2aXf6gwAAABgeyJvcmlnaW4iOiAiaHR0cHM6Ly9wd2EtZGlyZWN0b3J5LmFwcHNwb3QuY29tOjQ0MyIsICJmZWF0dXJlIjogIldlYlNoYXJlIiwgImV4cGlyeSI6IDE0ODEwNDQyMDJ9">
+
 <script id="config" type="application/json">
 {{{configstring}}}
 </script>
 <script src="/js/gulliver.js" async defer></script>
+<script type="text/javascript" src="https://www.gstatic.com/charts/loader.js" defer></script>
```
juliantoledo: I cannot change it, because it needs the order to load Google charts before running the script. How could we fix that?

lib/lighthouse.js
juliantoledo: done.

controllers/api.js
juliantoledo: I agree.

public/sw.js
```diff
@@ -13,6 +13,10 @@ const OFFLINE_URL = '/.shell/offline';
 
 const OFFLINE = [
   OFFLINE_URL,
+  '/img/GitHub-Mark-Light-32px.png',
```
juliantoledo: I cannot fix it, I did something wrong before.

PR #238, opened 2016-10-28
Add fetch() polyfill

ithinkihaveacat:
I think this is right now, although I don't know how to test—all the browsers I have support both `fetch()` and `Promise`.

### DIFF
--- a/public/js/gulliver.es6.js
+++ b/public/js/gulliver.es6.js
@@ -5,6 +5,11 @@
 
 /* eslint-env browser */
 
+// As used by https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Promise/config.json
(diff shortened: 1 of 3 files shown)


public/js/gulliver.es6.js
```diff
@@ -5,6 +5,7 @@
 
 /* eslint-env browser */
 
+import '../../node_modules/whatwg-fetch/fetch.js';
```
ithinkihaveacat: Yeah, it looks a bit funny because our build process is weird (the `*.es6.js` files don't need to be in the `public` directory): `rollup` merges all the modules together into `gulliver.js`, which is the only file that's needed in production. (Well, `lighthouse-chart.js` is needed too, but we'll move that eventually.)

PR #242, opened 2016-10-31
Add PWA using its homepage URL & Add encodedUrl field

juliantoledo:
We currently only support adding PWAs from it's manifest URL, this PR allows PWAs to be added by them homepage URL.


ithinkihaveacat:
Some PWAs, such as <https://www.pureformulas.com/>, have their manifests dynamically injected via JS. I doubt this is common, and they can still be added by explicitly passing the manifest URL, but if there's a nice way of working around this it's probably worth doing. (Can't think of one though, unless Lighthouse can extract for us.)

andreban:
Interesting. I can't see the Manifest information on Chrome Developer Tools. 
screen shot 2016-11-01 at 08 26 55

juliantoledo:
I could not find the manifest for pureformulas.com either. At the moment, I am only reading the html value.

juliantoledo:
@ithinkihaveacat @andreban PTAL

juliantoledo:
We can create another PR for handling the JS case, do you have any examples?

juliantoledo:
I added the Slugified field generation, I will create another PR to start using the field.

lib/pwa.js
```diff
@@ -211,6 +211,31 @@ exports.findByManifestUrl = function(manifestUrl) {
 };
 
 /**
+ * Finds a PWA by its slugified from DB.
+ *
+ * @param {string} slugified of the PWA's manifest
+ * @return {Pwa|null} the PWA from DB, or null if not found
+ */
+exports.findBySlugified = function(slugified) {
```
sebastianbenz: Is this still used?

test/lib/pwa.js
```diff
@@ -157,31 +158,56 @@ describe('lib.pwa', () => {
...
+      });
+    });
+  });
+
   describe('#save (core logic)', () => {
     afterEach(() => {
       simpleMock.restore();
     });
     it('performs all the save steps', () => {
       // Mock findByManifestUrl and bd to avoid making real calls
```
sebastianbenz: Your test setup requires a lot of mocks. This potentially makes your test very fragile. Too many mocks in a test is a sign that your method might be doing too much (and this why test are great - they can tell you a lot about your design). 

If I look at the name of the method you're testing, it's "save". However, your mocks tell me it's also updating the icon, updating lighthouse info and more. This a lot more than just save. Consider refactoring into a more pipeline based approach, e.g.:

```
const manifestProvider = new ManifestProvider(queryManifestFromDB, fetchManifest);
const manifestProcessors = [updatePwaIcon, updatePwaCache, updatePwaLighthouseInfo, savePwaToDb,...];

const manifestPipeline = new ManifestPipeline(manifestProvider, manifestProcessors);

manifestPipeline.execute(pwa);
```

ManifestPipeline.execute(...) could look something like this:

```
execute(pwa) {
  this._manifestProvider.get(pwa)
    .then(manifest =>  {
       pwa.manifest = manifest;
       return Promise.all(manifestProcessors.map(p => p.apply(pwa)));
    })
}
```

You can then test each ManifestProcessor in isolation and mock one or two ManifestProcessors for ManifestPipeline tests (without relying on any implementation details besides the ManifestProcessor contract). 

This has the positive side-effect that you no longer need to export methods for testing (which is your test trying to tell you that your method/class is doing too much).

juliantoledo: @sebastianbenz great feedback, we keep postponing refactoring this. I like your pipeline base approach, I will take a look into it :)

juliantoledo:
I will create another PR for further refactoring of Save and its test later.

lib/data-fetcher.js
```diff
@@ -42,14 +33,30 @@ const USER_AGENT = ['Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) Appl
  */
 function fetchMetadataDescription(url) {
   return fetchWithUA(url)
-    .then(response => DOMPurify.sanitize(response.text()))
```
amedina: No need to sanitize the response?

juliantoledo: @amedina The issue is that it breaks the parsing and the description was empty.

PR #243, opened 2016-11-02
Adds entity-counts to Gulliver

andreban:
Entity counts are added using a transaction and an entity with a
special key `entity-count` in the entity.


andreban:
Test deployed here: https://gulliver-1371.appspot.com/

juliantoledo:
@andreban It's looking good, you could use similar CSS to ".pwabox-right .name" to make it match the PWSs titles or the buttons text size and weight.

andreban:
@juliantoledo Updated the style. Deployed on the same server.

juliantoledo:
@andreban I haven't been able to test it because the error I mentioned over chat. I will try again later.

andreban:
@juliantoledo I found the problem. Can you test again?

juliantoledo:
LGTM

andreban:
@juliantoledo I've squashed the commits. Let's chat before merging / deploying so we can manually set the count for existing apps.

juliantoledo:
tested, deployed and count updated.

PR #245, opened 2016-11-09
Fix date problem in lighthouse test

juliantoledo:


juliantoledo:
@andreban small change in one test broken in different time zones.

PR #246, opened 2016-11-11
Track offline and homescreen traffic on Analytics

andreban:
- Adds `?utm-source` to  start_url on manifest, which will allow to identify clicks coming from the home-screen.
- Adds custom-dimension to GA that adds `navigator.onLine` as to pageview tracking.


juliantoledo:
LGTM

PR #247, opened 2016-11-11
Ensure start_url is cached and updated

andreban:
Updated sw.js to catch requests to start_url
(`/?utm_source=homescreen`) and respond with the root page (`/`).
This ensures that whenever the root page is updated, the latest
cached page will be served when launched from the homescreen.


andreban:
Deployed here: https://gulliver-1371.appspot.com/

juliantoledo:
LGTM

Issue #231, opened 2016-10-25
Cannot add twitter Error: Error: The value of property "raw" is longer than 1500 bytes..


kenchris:
https://mobile.twitter.com/manifest.json

juliantoledo:
@kenchris good catch. We should change the field type in the datastore.

juliantoledo:
Fixed, will be deployed in a couple of says.

PR #256, opened 2016-11-29
Manifest fixes

### LINKED ISSUE #255: Add support for Icons using Data URIs
Add support for Icons using Data URIs, for example:
https://mobile.twitter.com/manifest.json

Uses:
"icons":[{"src":"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA....

juliantoledo:
Fixes #255 and #231


juliantoledo:

lib/images.js
```diff
@@ -18,61 +18,75 @@
...
 function fetchAndSave(url, destFile) {
-  return new Promise((resolve, reject) => {
-    fetch(url)
-      .then(response => {
-        if (response.status !== 200) {
-          reject(
-            new Error('Bad Response (' + response.status + ') loading image: ' + response.url));
-        }
-        const contentType = response.headers.get('Content-Type');
+  if (url.startsWith('data:')) {
```
andreban: So, we currently support `http`, `https` and `data` protocol. We could write this as  
switch(protocol) {
  case 'data': {
    //...
  } 
  case 'http': 
  case 'https': {
    // ...
  }
  default: {
    return Promise.reject('Unsupported Protocol: ' + protocol);
  }
}

lib/images.js
```diff
@@ -18,61 +18,75 @@
...
+    return this.dataUriAndSave(url);
+  }
+  return dataFetcher.fetchWithUA(url)
+    .then(response => {
+      if (response.status !== 200) {
+        return Promise.reject(new Error(
+          'Bad Response (' + response.status + ') loading image: ' + response.url));
+      }
+      const contentType = response.headers['Content-Type'];
+      return Promise.all([
```
andreban: We ended up repeating the Promise.all here and in the function below. We could probably create a function with this and avoid having to change in 2 places if we decide to add another icon size, or change the sizes.

juliantoledo:
@andreban fixed comments, agree with the PR (I got carried away on the same one).

PR #257, opened 2016-12-01
Css for notifications

juliantoledo:


juliantoledo:
Let's merge and commit this after the other UI fixes going on.

PR #258, opened 2016-12-01
Reorganize api controllers

andreban:
Created an '/controllers/api' folder to make it easier to add more
API's without putting all off them inside api.js. This is necessary
for the upcoming addition of the notification subscription API.

### DIFF
--- a/controllers/api/index.js
+++ b/controllers/api/index.js
@@ -16,7 +16,7 @@
 'use strict';
 
 const express = require('express');
-const router = express.Router(); // eslint-disable-line
+const router = express.Router(); // eslint-disable-line new-cap
 
 // Includes APIs for Lighthouse (/api/lighthouse)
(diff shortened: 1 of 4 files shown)


controllers/api/index.js
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const express = require('express');
+const router = express.Router(); // eslint-disable-line
```
ithinkihaveacat: Nit: can you make this `eslint-disable-line new-cap` to disable only the `new-cap` rule?

andreban:
Fixed nit.

PR #271, opened 2016-12-07
Adds push notifications to Gulliver

andreban:
- Adds subscription the `new-apps` topic.

### DIFF
--- a/controllers/api/notifications.js
+++ b/controllers/api/notifications.js
@@ -36,7 +36,6 @@ router.get('/topics/', (req, res) => {
 });
 
 router.post('/subscribe/:topic/', jsonParser, (req, res) => {
-  console.log(req.body);
   const token = req.body.token;
   const topic = req.params.topic;
   notificationsLib.subscribe(token, topic)
--- a/controllers/api/notifications.js
+++ b/controllers/api/notifications.js
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2015-2016, Google, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const express = require('express');
+const bodyParser = require('body-parser');
+const router = express.Router(); // eslint-disable-line new-cap
+const notificationsLib = require('../../lib/notifications');
+const jsonParser = bodyParser.json();
+
+router.get('/topics/', (req, res) => {
+  const token = req.query.token;
+  notificationsLib.list(token)
+    .then(subscriptions => {
+      res.json({
+        subscriptions: subscriptions
+      });
+    })
+    .catch(err => {
+      res.status(500);
+      res.json(err);
+    });
+});
+
+router.post('/subscribe/:topic/', jsonParser, (req, res) => {
+  console.log(req.body);
+  const token = req.body.token;
+  const topic = req.params.topic;
+  notificationsLib.subscribe(token, topic)
+    .then(_ => {
+      res.json({success: true});
+    })
+    .catch(err => {
+      res.status(500);
+      res.json(err);
+    });
+});
+
+router.post('/unsubscribe/:topic/', jsonParser, (req, res) => {
+  const token = req.body.token.trim();
+  const topic = req.params.topic;
+  notificationsLib.unsubscribe(token, topic)
+    .then(_ => {
+      res.json({success: true});
+    })
+    .catch(err => {
+      res.status(500);
+      res.json(err);
+    });
+});
+
+module.exports = router;
+
--- a/public/firebase-messaging-sw.js
+++ b/public/firebase-messaging-sw.js
@@ -21,10 +21,7 @@ importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
 const configUrl = '/messaging-config.json';
 caches.match(configUrl)
   .then(response => {
-    if (response) {
-      return response;
-    }
-    return fetch(configUrl);
+    return response || fetch(configUrl);
   })
   .then(response => {
     return response.json();
--- a/public/firebase-messaging-sw.js
+++ b/public/firebase-messaging-sw.js
@@ -18,7 +18,14 @@
 importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-app.js');
 importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
 
-fetch('/messaging-config.json')
+const configUrl = '/messaging-config.json';
+caches.match(configUrl)
+  .then(response => {
+    if (response) {
+      return response;
+    }
+    return fetch(configUrl);
+  })
   .then(response => {
     return response.json();
   })
@@ -31,4 +38,3 @@ fetch('/messaging-config.json')
       return self.registration.showNotification();
     });
   });
-
--- a/public/firebase-messaging-sw.js
+++ b/public/firebase-messaging-sw.js
@@ -18,12 +18,17 @@
 importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-app.js');
 importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
 
-firebase.initializeApp({
-  messagingSenderId: '653391209629'
-});
+fetch('/messaging-config.json')
+  .then(response => {
+    return response.json();
+  })
+  .then(config => {
+    firebase.initializeApp({
+      messagingSenderId: config.firebase_msg_sender_id
+    });
+    const messaging = firebase.messaging();
+    messaging.setBackgroundMessageHandler(_ => {
+      return self.registration.showNotification();
+    });
+  });
 
-const messaging = firebase.messaging();
-messaging.setBackgroundMessageHandler(payload => {
-  console.log(payload);
-  return self.registration.showNotification('New App');
-});
--- a/public/firebase-messaging-sw.js
+++ b/public/firebase-messaging-sw.js
@@ -20,12 +20,8 @@ importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
 
 const configUrl = '/messaging-config.json';
 caches.match(configUrl)
-  .then(response => {
-    return response || fetch(configUrl);
-  })
-  .then(response => {
-    return response.json();
-  })
+  .then(response => response || fetch(configUrl))
+  .then(response => response.json())
   .then(config => {
     firebase.initializeApp({
       messagingSenderId: config.firebase_msg_sender_id
--- a/public/firebase-messaging-sw.js
+++ b/public/firebase-messaging-sw.js
@@ -13,15 +13,17 @@
  * limitations under the License.
  */
 
-'use strict';
+/* eslint-env serviceworker, browser */
+/* global firebase */
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-app.js');
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
 
-const express = require('express');
-const router = express.Router(); // eslint-disable-line new-cap
+firebase.initializeApp({
+  messagingSenderId: '653391209629'
+});
 
-// Includes APIs for Lighthouse (/api/lighthouse)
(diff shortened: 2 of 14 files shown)


public/firebase-messaging-sw.js
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+/* eslint-env serviceworker, browser */
+/* global firebase */
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-app.js');
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
+
+firebase.initializeApp({
+  messagingSenderId: '653391209629'
```
sebastianbenz: Don't commit senderId. Use config instead.

public/manifest.json
```diff
@@ -17,5 +17,6 @@
 	],
 	"theme_color": "#7cc0ff",
 	"background_color": "#7cc0ff",
-	"display": "standalone"
+	"display": "standalone",
+	"gcm_sender_id": "103953800507"
```
sebastianbenz: Don't commit senderId. Use config instead.

public/js/messaging.js
```diff
@@ -0,0 +1,123 @@
...
+        return fetch(url);
+      })
+      .then(response => {
+        return response.json();
+      })
+      .then(json => {
+        return json.subscriptions;
+      })
+      .catch(_ => {
+        return Promise.resolve([]);
```
sebastianbenz: Consider adding some logging for the error case.

public/js/gulliver.es6.js
```diff
@@ -278,13 +278,38 @@ function setupConfig() {
...
 
+function setupMessaging() {
+  const messaging = new Messaging();
+  messaging.init();
+
+  const checkbox = document.getElementById('notifications');
+  if (checkbox) {
+    checkbox.addEventListener('change', e => {
+      if (e.target.checked) {
+        messaging.subscribe('new-apps')
```
sebastianbenz: extract `new-apps` into a constant

public/js/gulliver.es6.js
```diff
@@ -278,13 +278,38 @@ function setupConfig() {
   }
 }
 
+function setupMessaging() {
+  const messaging = new Messaging();
+  messaging.init();
+
+  const checkbox = document.getElementById('notifications');
```
sebastianbenz: It might make sense to encapsulate the notification element handler into a separate class (maybe even file).

sebastianbenz: A good way to propagate the notification state to the element handler could be registering the element handler as a listener to messaging.js. Messaging.js would fire events for status changes such as: permission denied, permission allowed, ...

public/manifest.json
andreban: `gcm_sender_id` is the same for ALL Firebase Messaging projects. I'm inclined to leave this as a constant since it won't change for different deploys. Check info here:
https://firebase.google.com/docs/cloud-messaging/js/client

ithinkihaveacat: Maybe add a comment similar to https://firebase.google.com/docs/cloud-messaging/js/client#configure_the_browser_to_receive_messages?

andreban: sounds good

public/firebase-messaging-sw.js
andreban: Yes.

public/js/messaging.js
andreban: Will add logging. We need to add a component to display Toasts and communicate errors to users. Created an issue for the Toasts on #277

public/firebase-messaging-sw.js
andreban: done.

public/js/gulliver.es6.js
andreban: Created a NotificationCheckbox class on a separate file. I think we should start discussing using web-components in this scenario: we could encapsulate the Html, CSS and JS in a single file.

controllers/api/notifications.js
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const express = require('express');
+const bodyParser = require('body-parser');
+const router = express.Router(); // eslint-disable-line new-cap
+const notificationsLib = require('../../lib/notifications');
```
sebastianbenz: nit: can you find a better name that notificationsLib?

public/firebase-messaging-sw.js
```diff
@@ -0,0 +1,40 @@
...
+
+/* eslint-env serviceworker, browser */
+/* global firebase */
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-app.js');
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
+
+const configUrl = '/messaging-config.json';
+caches.match(configUrl)
+  .then(response => {
+    if (response) {
```
sebastianbenz: shorter:

```
return response ||  fetch(configUrl);
```

public/js/gulliver.es6.js
```diff
@@ -217,10 +218,9 @@ function setupSignin() {
     });
 
     const authButton = document.getElementById('auth-button');
-    const authButtonLabel = document.getElementById('auth-button-label');
 
     function updateAuthButtonLabel() {
-      authButtonLabel.innerText = authButton.dataset.signedin === 'true' ?
+      authButton.innerText = authButton.dataset.signedin === 'true' ?
```
sebastianbenz: ups - I broke this. Thanks for fixing.

public/js/messaging.js
```diff
@@ -0,0 +1,124 @@
...
+      .then(token => {
+        console.log(token);
+        const url = SUBSCRIBE_ENDPOINT + '/' + topic;
+        return fetch(url, {
+          method: 'POST',
+          headers: {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json'
+          },
+          body: JSON.stringify({token: token})
```
sebastianbenz: extract: 

```
return fetch(url, {
 +          method: 'POST',
 +          headers: {
 +            'Accept': 'application/json',
 +            'Content-Type': 'application/json'
 +          },
 +          body: JSON.stringify({token: token})
 +        });
```

into a separate method and use for subscribe and unsubscribe

public/js/messaging.js
```diff
@@ -0,0 +1,124 @@
...
+   * @returns Promise<Array<String>>
+   */
+  getSubscriptions() {
+    const messaging = firebase.messaging();
+    return messaging.getToken()
+      .then(token => {
+        if (!token) {
+          return Promise.resolve([]);
+        }
+        const url = TOPICS_ENDPOINT + '?token=' + token;
```
sebastianbenz: Don't we need to set an accept header?

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+import Messaging from './messaging';
+
+export default class NotificationCheckbox {
+  init(firebaseMsgSenderId, checkbox, topic) {
```
sebastianbenz: Why not create a constructor for this?

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,46 @@
...
+    if (!checkbox) {
+      return;
+    }
+    this.messaging = new Messaging();
```
sebastianbenz: This feels like it doesn't belong inside this class, but should be passed in as a dependency via the constructor. Multiple classes might need to access messaging...

public/js/messaging.js
```diff
@@ -0,0 +1,124 @@
...
+import 'whatwg-fetch/fetch'; // Imports `fetch` polyfill
+import firebase from 'firebase/app';
+import 'firebase/messaging';
+
+const SUBSCRIBE_ENDPOINT = '/api/notifications/subscribe';
+const UNSUBSCRIBE_ENDPOINT = '/api/notifications/unsubscribe';
+const TOPICS_ENDPOINT = '/api/notifications/topics';
+
+export default class Messaging {
+  init(messagingSenderId) {
```
sebastianbenz: Why not do this in a constructor?

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,46 @@
...
+    this.messaging.init(firebaseMsgSenderId);
+    if (checkbox) {
```
sebastianbenz: this is always true as you've already checked earlier

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,46 @@
...
+      checkbox.addEventListener('change', e => {
+        if (e.target.checked) {
+          this.messaging.subscribe(topic)
+            .catch(e => {
+              console.error('Error subscribing to topic: ', e);
+              e.target.checked = false;
```
sebastianbenz: What are reasons for this to fail? What if the user doesn't give the permission. Should we provide a separate UI for this case?

controllers/api/notifications.js
andreban: This is the same pattern that is being used on the other controllers (pwaLib, lighthouseLib, etc). I'd rather keep for now and if it makes sense, change on all controllers.

public/js/notification-checkbox.js
andreban: I think we need a separate UI. I've created #277 so we can display a Toast when subscribing fails, so we can notify the user.

controllers/api/notifications.js
sebastianbenz: ok

public/js/notification-checkbox.js
sebastianbenz: This is a good sample for how to handle the UI part: https://gauntface.github.io/simple-push-demo/

sebastianbenz: I would simply disable the checkbox in this case.

andreban: The user may have forgotten he has blocked notifications. We need to communicate that he needs to re-enable it in the browser configuration if he wants to enable notifications.

public/js/notification-checkbox.js
andreban: done

public/js/notification-checkbox.js
andreban: done.

public/js/notification-checkbox.js
andreban: done.

public/js/messaging.js
andreban: Not really, but added as it is a best practice.

public/js/messaging.js
andreban: done

public/firebase-messaging-sw.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+/* eslint-env serviceworker, browser */
+/* global firebase */
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-app.js');
+importScripts('https://www.gstatic.com/firebasejs/3.5.2/firebase-messaging.js');
+
+const configUrl = '/messaging-config.json';
+caches.match(configUrl)
+  .then(response => {
```
sebastianbenz: You could write this even shorter:

```
.then(response =>  response || fetch(configUrl) )
.then(response => response.json()(
```

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class NotificationCheckbox {
+  constructor(messaging, checkbox, topic) {
+    if (!checkbox) {
```
sebastianbenz: log an error

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,41 @@
...
+      return;
+    }
+    this.messaging = messaging;
+    checkbox.addEventListener('change', e => {
```
sebastianbenz: Moving the event handler into a separate method makes it a bit more readable (and testable)

public/js/notification-checkbox.js
```diff
@@ -0,0 +1,41 @@
...
+      if (e.target.checked) {
+        this.messaging.subscribe(topic)
+          .catch(e => {
+            console.error('Error subscribing to topic: ', e);
+            e.target.checked = false;
+          });
+        return;
+      }
+      this.messaging.unsubscribe(topic);
```
sebastianbenz: Can unsubscribe not fail?

sebastianbenz:
Added a few more comments. Mostly nits.

andreban:
Latest version deployed here: https://gulliver-1371.appspot.com/

public/js/notification-checkbox.js
andreban: done

public/js/notification-checkbox.js
andreban: done

PR #272, opened 2016-12-07
Improve config error detection handling

ithinkihaveacat:
Try to be a bit better at noticing and diagnosing configuration errors.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -26,7 +26,6 @@ nconf
   // 2. Environment variables
   .env([
     'CLOUD_BUCKET',
-    'DATA_BACKEND',
     'GCLOUD_PROJECT',
     'PORT',
     'CLIENT_ID',
@@ -41,7 +40,6 @@ nconf
   .file({file: path.join(__dirname, 'config.json')})
   // 4. Defaults
   .defaults({
-    DATA_BACKEND: 'datastore',
     PORT: 8080 // Port used by HTTP server
   });
 
--- a/config/config.js
+++ b/config/config.js
@@ -54,9 +54,9 @@ function checkConfig(setting) {
   if (!nconf.get(setting)) {
     throw new Error(`You must set the ${setting} environment variable or add it to config.json!`);
   }
-  // If setting includes a space, emit warning
+  // If setting includes a space, throw error
   if (nconf.get(setting).match(/\s/)) {
-    console.warn(`The ${setting} environment variable is suspicious ("${nconf.get(setting)}")`);
+    throw new Error(`The ${setting} environment variable is suspicious ("${nconf.get(setting)}")`);
   }
 }
 
--- a/config/config.js
+++ b/config/config.js
@@ -17,7 +17,7 @@
 
 // Hierarchical node.js configuration with command-line arguments, environment
 // variables, and files.
-const nconf = module.exports = require('nconf');
+const nconf = require('nconf');
 const path = require('path');
 
 nconf
@@ -41,15 +41,8 @@ nconf
   .file({file: path.join(__dirname, 'config.json')})
   // 4. Defaults
   .defaults({
-    // Typically you will create a bucket with the same name as your project ID.
-    CLOUD_BUCKET: '',
     DATA_BACKEND: 'datastore',
-
-    // This is the id of your project in the Google Cloud Developers Console.
-    GCLOUD_PROJECT: '',
-
-    // Port the HTTP server
-    PORT: 8080
+    PORT: 8080 // Port used by HTTP server
   });
 
 // Check for required settings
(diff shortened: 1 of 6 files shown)


config/config.js
```diff
@@ -17,7 +17,7 @@
 
 // Hierarchical node.js configuration with command-line arguments, environment
 // variables, and files.
-const nconf = module.exports = require('nconf');
```
ithinkihaveacat: Seems to be more conventional (at least in our project) to do this at the bottom of the file.

config/config.js
```diff
@@ -39,15 +39,8 @@ nconf
   .file({file: path.join(__dirname, 'config.json')})
   // 4. Defaults
   .defaults({
-    // Typically you will create a bucket with the same name as your project ID.
-    CLOUD_BUCKET: '',
```
ithinkihaveacat: Thing we should just pull the defaults from `config.json` (where they're now sort of documented) instead of having configuration & documentation in multiple places.

config/config.json
```diff
@@ -1,7 +1,7 @@
 {
-  "GCLOUD_PROJECT": "google.com:gulliver",
-  "CLOUD_BUCKET": "gulliver.google.com.a.appspot.com",
-  "CLIENT_ID": "605287872172-js6omne47i1k79hnfo7d4bdu9rlemslr.apps.googleusercontent.com",
+  "GCLOUD_PROJECT": "run `gcloud config get-value project`",
+  "CLOUD_BUCKET": "your cloud bucket",
```
sebastianbenz: can you add a sample? I for example didn't know that this'd be the normal URL

config/config.json
```diff
@@ -1,7 +1,7 @@
...
+  "CLIENT_ID": "your client id",
```
sebastianbenz: Can your add a link to where to find it

config/config.js
```diff
@@ -57,8 +50,14 @@ checkConfig('CLIENT_ID');
...
 function checkConfig(setting) {
+  // If setting undefined, throw error
   if (!nconf.get(setting)) {
-    throw new Error('You must set the ' + setting + ' environment variable or' +
-      ' add it to config.json!');
+    throw new Error(`You must set the ${setting} environment variable or add it to config.json!`);
+  }
+  // If setting includes a space, emit warning
+  if (nconf.get(setting).match(/\s/)) {
+    console.warn(`The ${setting} environment variable is suspicious ("${nconf.get(setting)}")`);
```
sebastianbenz: You could trim before checking to ignore trailing whitespace

config/config.json
ithinkihaveacat: Yeah good point, was thinking there was no concise way of explaining this but maybe `see https://console.cloud.google.com/storage/browser?project=$GCLOUD_PROJECT` is an improvement?

config/config.js
ithinkihaveacat: Possibly, although I think whitespace at the start or end is at least suspicious?

config/config.json
```diff
@@ -1,10 +1,11 @@
 {
-  "GCLOUD_PROJECT": "google.com:gulliver",
-  "CLOUD_BUCKET": "gulliver.google.com.a.appspot.com",
-  "CLIENT_ID": "605287872172-js6omne47i1k79hnfo7d4bdu9rlemslr.apps.googleusercontent.com",
-  "CLIENT_SECRET": "your client secret",
+  "//": "See README.md for more information about what to put here",
+  "GCLOUD_PROJECT": "run `gcloud config get-value project`",
+  "CLOUD_BUCKET": "see https://console.cloud.google.com/storage/browser?project=$GCLOUD_PROJECT",
+  "CLIENT_ID": "see https://console.cloud.google.com/apis/credentials?project=$GCLOUD_PROJECT",
```
ithinkihaveacat: @hostirosti Is there a way to extract these values using the `gcloud` tool, instead of directing users to a URL for them to extract "manually"? (i.e. for a given project, return associated cloud buckets, client_ids, etc.) I suppose some sort of script could do it, but I don't want to go overboard with automation...

config/config.js
sebastianbenz: ...or sloppy ;-). If in doubt - throw an error.

config/config.js
```diff
@@ -39,15 +39,8 @@ nconf
   .file({file: path.join(__dirname, 'config.json')})
   // 4. Defaults
   .defaults({
-    // Typically you will create a bucket with the same name as your project ID.
-    CLOUD_BUCKET: '',
     DATA_BACKEND: 'datastore',
```
juliantoledo: We could remove it, it was in the original sample, we just need to remove it in few places where it is used to build model-datastore lib name.

ithinkihaveacat: 👍 Removed in 8703a72.

config/config.js
ithinkihaveacat: Upgraded to error in 2fb2d3a.

PR #276, opened 2016-12-08
Store raw lighthouse report

juliantoledo:
Store raw lighthouse report


models/lighthouse.js
```diff
@@ -25,15 +27,24 @@
...
 class Lighthouse {
-  constructor(pwaId, absoluteStartUrl, lighthouseInfo) {
+  constructor(pwaId, absoluteStartUrl, lighthouseJson) {
     this.pwaId = pwaId;
     this.absoluteStartUrl = absoluteStartUrl;
-    this.lighthouseInfo = lighthouseInfo;
-    this.totalScore = lighthouseInfo.totalScore;
-    this.lighthouseVersion = lighthouseInfo.lighthouseVersion;
+    this.lighthouseInfo = libLighthouse.processLighthouseJson(lighthouseJson);
+    this._lighthouseJson = JSON.stringify(lighthouseJson);
```
sebastianbenz: shouldn't this be JSON.parse(lighthouseJson)?

sebastianbenz: Or if it is already parsed then simply store the json object.

juliantoledo: We need to store the raw string, I tried the object and that does not fit the DB field because it needs to be a string to be bigger than 1500 bytes.

juliantoledo: I can change the constructor to receive just the string and avoid parsing twice.

juliantoledo: Actually, it is a JSON object everywhere, so I rather keep encapsulated the conversion to string within the model.

juliantoledo:
@sebastianbenz  We need to store the raw string, I tried the object and that does not fit the DB field because it needs to be a string to be bigger than 1500 bytes. It is a JSON object everywhere, so I rather keep encapsulated the conversion to string within the model.

sebastianbenz:
I'd prefer to keep the json object in the model. The json object is potentially read multiple times and parsing it into a json object everytime is expensive. The string is only used once - when saving.

juliantoledo:
Sounds good, I will take a look at changing model-datastore to perform the conversion when saving/loading.

juliantoledo:
@sebastianbenz PTAL

models/lighthouse.js
```diff
@@ -15,25 +15,41 @@
...
-    this.totalScore = lighthouseInfo.totalScore;
-    this.lighthouseVersion = lighthouseInfo.lighthouseVersion;
+    this._lighthouseJson = parseToJson(lighthouseJson);
+    this.lighthouseInfo = libLighthouse.processLighthouseJson(this._lighthouseJson);
+    this.totalScore = this.lighthouseInfo.totalScore;
+    this.lighthouseVersion = this.lighthouseInfo.lighthouseVersion;
     this.date = (new Date()).toISOString().slice(0, 10);
     this.id = this.pwaId + '-' + this.date;
   }
+
```
sebastianbenz: The getter & setter are never used -> I'd prefer to remove them until we actually need them

models/lighthouse.js
```diff
@@ -15,25 +15,34 @@
...
+  get lighthouseJson() {
+    return JSON.parse(this._lighthouseJson);
+  }
+
+  set lighthouseJson(value) {
```
juliantoledo: It is used when we read from DB, during Object.assigns

models/lighthouse.js
juliantoledo: We need them for when we read from DB.

sebastianbenz:
LGTM then

PR #279, opened 2016-12-08
Minify rendered HTML

andreban:
- Add express-minify-html and configure it in app.js


app.js
```diff
@@ -45,6 +46,19 @@ app.locals.configstring = JSON.stringify({
 
 app.use(bodyParser.urlencoded({extended: true}));
 
+app.use(minifyHTML({
```
sebastianbenz: Should we do this only in prod?

andreban: done.

PR #287, opened 2016-12-13
Send push notification when a new PWA is added

andreban:
- Hook the push event after a new app is saved on lib/pwa


andreban:
@juliantoledo @sebastianbenz PTAL

firebaseMsgSwGenerator.js
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const fs = require('fs');
+const template = require('lodash.template');
```
sebastianbenz: Why introduce a new templating engine?

lib/pwa.js
```diff
@@ -323,10 +324,14 @@ exports.fetchManifest = function(pwa) {
  * @return {Pwa} the updated PWA
  */
 exports._save = function(pwa) {
+  let created = true;
   return this.fetchManifest(pwa)
   .then(manifest => {
```
sebastianbenz: nit: formatting

lib/pwa.js
```diff
@@ -323,10 +324,14 @@ exports.fetchManifest = function(pwa) {
...
     return this.findByManifestUrl(pwa.manifestUrl)
     .then(existingPwa => {
```
sebastianbenz: nit: formatting

lib/pwa.js
```diff
@@ -323,10 +324,14 @@ exports.fetchManifest = function(pwa) {
  * @return {Pwa} the updated PWA
  */
 exports._save = function(pwa) {
+  let created = true;
```
sebastianbenz: Not sure I like storing the created state in a variable in the outer scope. Maybe return a wrapper object inside the promise instead, e.g.:

```
return {
  pwa: pwa,
  created: true
};
```

firebaseMsgSwGenerator.js
andreban: Handlebars uses braces (`{{` and `}}`), as a delimiter for it's templates, and it may conflict with JS braces. It seems it's not possible to change delimiters in handlebars (http://stackoverflow.com/questions/14324850/how-to-change-the-default-delimiter-of-handlebars-js). Also, `lodash.template` is already being indirectly included in the project by `sw-precache`, as it uses the same templating engine to build the service-worker.

sebastianbenz:
Good point. :-)

firebaseMsgSwGenerator.js
sebastianbenz: True - haven't thought of that.

firebaseMsgSwGenerator.js
```diff
@@ -0,0 +1,39 @@
+/**
```
ithinkihaveacat: Can the name more closely match the template and generated response? Maybe `firebase-messaging-sw-generator.js`? Another option would be to go more generic, so that it works on any similar `.tmpl` file.

firebase-messaging-sw.tmpl
```diff
@@ -18,16 +18,10 @@
...
-    firebase.initializeApp({
-      messagingSenderId: config.firebase_msg_sender_id
-    });
-    const messaging = firebase.messaging();
-    messaging.setBackgroundMessageHandler(_ => {
-      return self.registration.showNotification();
-    });
-  });
+firebase.initializeApp({
+  messagingSenderId: '<%= firebaseMsgSenderId %>'
```
ithinkihaveacat: Might be better to have the injected config in one place (just this one key at the moment), and reference it here via something like `__config.firebaseMsgSenderId`. (Similar to the way we reference `__config` from the page JS.) Not that bothered though--could wait for another key to be added. Generating code is all kinds of icky…

firebaseMsgSwGenerator.js
andreban: I renamed the generator. I think we won't need more generators anytime soon, so just keeping the specific one instead of making it more generic. If another one is added, we should probably make it generic.

config/config.js
```diff
@@ -33,7 +33,8 @@ nconf
     'CLIENT_SECRET',
     'LIGHTHOUSE_SERVER',
     'GOOGLE_ANALYTICS',
-    'FIREBASE_AUTH'
+    'FIREBASE_AUTH',
+    'NOTIFICATION_DOMAIN'
```
andreban: Makes sense. Changed.

lib/pwa.js
andreban: Done.

andreban:
I don't like the current layout of the save method too. One thing that is bad on the design is that if a task, such as fetching lighthouse information or fetching images fails, there's nothing keeping track of it and it's not possible to automatically retry. Maybe we should use the Tasks framework @juliantoledo created and turn those into tasks. Another option would be to create a bus where Lighthouse, Notifications, Images could sign up for events on PWA's being added or updated and take control from there.

sebastianbenz:
+1 for EventBus. This would also solve our logging problem (log entries are just events).

juliantoledo:
+1 we really need to refactor it. I don't have experience with EventBus, but we should really consider it. I think we should rather do a new PR only with that.

juliantoledo:
Let's kickoff the save/update refactoring here: #297

Issue #282, opened 2016-12-12
Error: TypeError: pwa.manifest.getBestIconUrl is not a function.


PaulKinlan:
Tried to add https://m.snapdeal.com/manifest.json got Error: TypeError: pwa.manifest.getBestIconUrl is not a function. a quick glance of manifest and it looks ok.

![image](https://cloud.githubusercontent.com/assets/45510/21119881/82fb0ea4-c0bc-11e6-815f-e67803c35ab6.png)

juliantoledo:
I fixed our error handling for it, the problem is not the icon, is that the manifest is not valid because "Error: color parsing failed."

According to the lighthouse manifest-parser:
https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/lib/manifest-parser.js

PR #289, opened 2016-12-13
Improve error handling for #282

juliantoledo:
Reject manifest if it does not pass the Lighthouse manifest-parser validation and show the error from it.

I also updated the manifest-parser.

### DIFF
--- a/lib/pwa.js
+++ b/lib/pwa.js
@@ -336,8 +336,10 @@ exports._save = function(pwa) {
 
       // Validates the manifest or, reject and return errors in an array
       const res = libManifest.validateManifest(manifest.raw, pwa.manifestUrl, pwa.absoluteStartUrl);
-      pwa.manifest = res.length === 0 ? manifest : Promise.reject(res);
-
+      if (res.length !== 0) {
+        return Promise.reject('Error while validating the manifest: ' + res);
+      }
+      pwa.manifest = manifest;
       return this.savePwa(pwa);
     })
     .then(savedPwa => {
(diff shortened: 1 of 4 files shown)


juliantoledo:
PTAL

lib/pwa.js
```diff
@@ -336,8 +336,10 @@ exports._save = function(pwa) {
 
       // Validates the manifest or, reject and return errors in an array
       const res = libManifest.validateManifest(manifest.raw, pwa.manifestUrl, pwa.absoluteStartUrl);
-      pwa.manifest = res.length === 0 ? manifest : Promise.reject(res);
-
+      if (res.length !== 0) {
+        return Promise.reject(res);
```
juliantoledo: res contains the rejection information from the validator, I could add an additional message to it.

third_party/manifest-parser.js
```diff
@@ -157,7 +157,7 @@ function parseStartUrl(jsonInput, manifestUrl, documentUrl) {
 }
 
 function parseDisplay(jsonInput) {
-  let display = parseString(jsonInput.display, true);
+  const display = parseString(jsonInput.display, true);
```
ithinkihaveacat: @juliantoledo Did you use https://github.com/GoogleChrome/gulliver/blob/master/third_party/install.sh to generate this? Just wondering why the updated `Color.js` isn't here as well. (Which seems to include some awkward changes, but still probably better to commit both together.)

juliantoledo: @ithinkihaveacat I also updated our script to work with their latest version of both files.

PR #293, opened 2016-12-13
Log Firebase notification HTTP errors

ithinkihaveacat:
HTTP level errors get swallowed by the notification abstraction layer.

### DIFF
--- a/lib/data-fetcher.js
+++ b/lib/data-fetcher.js
@@ -157,8 +157,16 @@ function _handleFirebaseResponse(response) {
 
 function firebaseFetch(url, payload) {
   const options = _firebaseOptions(payload);
-  return fetch(url, options)
-    .then(_handleFirebaseResponse);
+  const res = fetch(url, options);
+  res.then(r => {
+    if (r.status !== 200) {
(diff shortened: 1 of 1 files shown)


lib/data-fetcher.js
```diff
@@ -157,8 +157,16 @@ function _handleFirebaseResponse(response) {
...
-    .then(_handleFirebaseResponse);
+  const res = fetch(url, options);
+  res.then(r => {
+    if (r.status !== 200) {
+      r.text().then(msg => {
+        // Add codebase-wide logging system
+        console.warn(`firebaseFetch error: GET ${url} => ${r.status}: ${msg}`);
+      });
+    }
+  });
```
sebastianbenz: Is there a particular reason why you don't chain the then calls?

ithinkihaveacat: I think not chaining the calls make it more clear that this is an observer that does not transform the response--if you're scanning through `firebaseFetch()` to see what it does/returns, you can ignore this block, because it's not chained, and not connected to the return value.

Issue #304, opened 2016-12-16
Adding Twitter Fails 


PaulKinlan:
It reports that the start Url is not a URL... It is.

{
  "name": "Twitter",
  "short_name": "Twitter",
  "start_url": "/?utm_source=web_app_manifest",
  "display": "standalone",
  "gcm_sender_id": "49625052041",
  "gcm_user_visible_only": true,
  "prefer_related_applications": true,
  "related_applications": [
    {
      "platform": "play",
      "id": "com.twitter.android"
    }
  ]
}

PR #306, opened 2016-12-19
Fixes a manifest validation and lets add PWAs without icon

juliantoledo:


juliantoledo:
Fixes #304

juliantoledo:
@ithinkihaveacat @sebastianbenz PTAL

sebastianbenz:
Can you update the tests as well?

juliantoledo:
Of course, I just added one manifest without icon, I will also add another test with a invalid manifest.

juliantoledo:
done

test/manifests/no-icon.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Twitter",
+  "short_name": "Twitter",
+  "start_url": "/?utm_source=web_app_manifest",
+  "display": "standalone",
+  "gcm_sender_id": "49625052041",
```
sebastianbenz: nit: to harden test cases against future changes, I'd recommend limiting test data to the minimum needed to expose a specific behavior. In this case this'd mean removing: gcm_sender_id, gcm_user_visible_only, related_applications,...

test/manifests/invalid.json
```diff
@@ -0,0 +1,7 @@
+{
```
sebastianbenz: nit: you could explain why this manifest is invalid via the filename (I actually don't know why this is invalid)

test/lib/pwa.js
```diff
@@ -39,23 +39,39 @@ const MANIFEST_URL = 'https://www.terra.com.br/manifest-br.json';
...
         pwa = new Pwa(MANIFEST_URL, manifest);
         pwa.id = '123456789';
         dataFetcher.readFile(LIGHTHOUSE_JSON_EXAMPLE)
           .then(data => {
             lighthouse = new Lighthouse('123456789', 'www.domain.com', JSON.parse(data));
-            done();
           });
-      });
+      })],
+      [dataFetcher.readFile(MANIFEST_DATA_NO_ICON)
```
sebastianbenz: Is it possible to directly create the PWA obj? The advantages would be:

* faster tests
* more maintainable tests as they have less dependencies to other APIs (e.g. dataFetcher)

test/manifests/no-icon.json
juliantoledo: sure, sounds good.

test/lib/pwa.js
juliantoledo: I think so, I had a hard time escaping manifest properly to embed them in the test. I will try again.

juliantoledo: @sebastianbenz added the manifest as inline text and replaced the datafetcher for a regular fs call to read the lighthouse json.

juliantoledo:
PTAL

test/lib/pwa.js
```diff
@@ -13,50 +13,61 @@
...
-        pwa = new Pwa(MANIFEST_URL, manifest);
-        pwa.id = '123456789';
-        dataFetcher.readFile(LIGHTHOUSE_JSON_EXAMPLE)
-          .then(data => {
-            lighthouse = new Lighthouse('123456789', 'www.domain.com', JSON.parse(data));
-            done();
-          });
-      });
-  });
+  const manifest = new Manifest(MANIFEST_URL, JSON.parse(MANIFEST_DATA));
```
sebastianbenz: You can avoid the parse step if you don't declare the manifest as a string:

```
const MANIFEST_DATA = {
  name: "Test",
  ...
}

```

test/lib/pwa.js
```diff
@@ -13,50 +13,61 @@
...
+  const pwa = new Pwa(MANIFEST_URL, manifest);
+  pwa.id = '123456789';
+  const manifestNoIcon = new Manifest(MANIFEST_URL, JSON.parse(MANIFEST_NO_ICON));
+  const pwaNoIcon = new Pwa(MANIFEST_URL, manifestNoIcon);
```
sebastianbenz: I've created a helper method for instantiating the PWAs [here](https://github.com/GoogleChrome/gulliver/blob/master/test/models/pwa.js#L129). We could use this everytime we create a new pwa in a test. The benefit is that we only need to touch the helper method if for example the PWA constructor API changes.

test/lib/pwa.js
juliantoledo: I wanted to have the text as similar as possible to the real manifest, that is why I was reading them from files.

juliantoledo:
Fixed the latest comments.
@sebastianbenz PTAL

PR #324, opened 2017-01-09
Remove mocha reporter selection

### LINKED ISSUE #308: npm run mocha fails
```
npm run mocha

> gulliver@1.0.0 mocha /Users/sbenz/Documents/projects/gulliver
> _mocha -R  test/**/*

"test/controllers/api.js" reporter blew up with error:
ReferenceError: describe is not defined
    at Object.<anonymous> (/Users/sbenz/Documents/projects/gulliver/test/controllers/api.js:32:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:
...
```

ithinkihaveacat:
The `-R` switch seems to be bogus. Not sure why it worked when run under `istanbul` (maybe it was filtering it out?) but I don't think `npm run mocha` ever worked.

Also, I discovered that `npm` makes the binaries in `node_modules/.bin` available in the PATH, so the directory doesn't actually need to be hardcoded into `package.json`.

Fixes #308.


andreban:
Curiosity: what's the different between running `mocha` and `_mocha`?

ithinkihaveacat:
@andreban I wondered about that too. There's a [brief comment](https://github.com/mochajs/mocha/blob/master/bin/mocha#L5) at the top of `mocha` that is moderately helpful. (Not sure why we're using `_mocha` instead of `mocha`; both seem to work.)

PR #329, opened 2017-01-25
Favorite PWAs for user

juliantoledo:
Backend API and entities to store Favorite PWAs, I think we should use another PR for the frontend to avoid making this one bigger.


juliantoledo:
@andreban PTAL

juliantoledo:
@ithinkihaveacat @sebastianbenz PTAL

controllers/api/favorite-pwa.js
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * GET /favorite-pwa/
+ *
+ * Returns all Favorite PWAs for a user
+ */
+router.get('/', (req, res) => {
+  res.setHeader('Content-Type', 'application/json');
+
+  const idToken = req.query.idToken;
```
juliantoledo: We send it as part of the form data in Add PWA.
How would the client send it in this case? Do I get id_token as part of the Authorization header?

controllers/api/favorite-pwa.js
```diff
@@ -0,0 +1,124 @@
...
+      if (favoritePwa) {
+        res.json(favoritePwa);
+      } else {
+        res.status(404);
+        res.json('not found');
+      }
+    })
+    .catch(err => {
+      res.status(500);
+      res.json(err);
```
juliantoledo: done.

controllers/api/favorite-pwa.js
```diff
@@ -0,0 +1,124 @@
...
+ * GET /favorite-pwa/
+ *
+ * Returns all Favorite PWAs for a user
+ */
+router.get('/', (req, res) => {
+  res.setHeader('Content-Type', 'application/json');
+
+  const idToken = req.query.idToken;
+  if (!idToken) {
+    res.status(403);
```
juliantoledo: done.

lib/favorite-pwa.js
```diff
@@ -0,0 +1,90 @@
...
+const ENTITY_NAME = 'FAVORITE-PWA';
+
+/**
+ * Saves a FavoritePwa object into the DB.
+ *
+ * @param {FavoritePwa} lighthouse
+ * @return {Promise<FavoritePwa>}
+ */
+exports.save = function(favoritePwa) {
+  return new Promise((resolve, reject) => {
```
juliantoledo: done

controllers/api/favorite-pwa.js
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Create a Favorite PWA.
+ */
+router.post('/add', (req, res) => {
+  res.setHeader('Content-Type', 'application/json');
+  const idToken = req.body.idToken;
+  const isFavorite = req.body.isFavorite;
+  const pwaId = req.body.pwaId;
+
+  verifyIdToken.verifyIdToken(idToken)
```
juliantoledo: Thanks, I will look into it. At the moment we only need the token to Add and to handle favorites.

controllers/api/favorite-pwa.js
ithinkihaveacat: Sending the token as a POST parameter (as in "Add PWA") is better than a GET query string. (If it's in the query string, it's susceptible to being logged, saved to browser history, leaked to third parties via the referrer, etc.)

I think the way the "star" function should work is via an XHR/fetch() POST to some endpoint.  Currently the code [automatically sets the `idToken` form element](https://github.com/GoogleChrome/gulliver/blob/master/public/js/gulliver.es6.js#L78) when a page is loaded, or whenever the user changes. (If they switch accounts, for example.)

It could be enhanced to set `document.idToken` or similar for easy access to the starring code. (Or maybe create an `authenticatedFetch()` function that works like `fetch()`, except that it injects the `Authorization` header if available. This would also be a good point to add CSRF handling, if we do that.)

controllers/api/favorite-pwa.js
```diff
@@ -0,0 +1,124 @@
...
+      res.json(err);
+    });
+});
+
+/**
+ * POST /favorite-pwa/add
+ *
+ * Create a Favorite PWA.
+ */
+router.post('/add', (req, res) => {
```
juliantoledo: I will remove /add and leave / with POST to add and DELETE to delete.

lib/favorite-pwa.js
```diff
@@ -0,0 +1,90 @@
...
+    return favoritePwas;
+  });
+};
+
+/**
+ * Retrieves FavoritePwas for given User & PWA.
+ *
+ * @param {number} pwaId
+ * @param {number} userId
+ * @return {Promise<FavoritePwa>}
```
juliantoledo: A single one, done.

lib/favorite-pwa.js
```diff
@@ -0,0 +1,90 @@
...
+        return reject('Error saving the FavoritePwa');
+      });
+  });
+};
+
+/**
+ * Retrieves FavoritePwas for a given User.
+ *
+ * @param {number} userId
+ * @return {Promise<FavoritePwa>}
```
juliantoledo: done.

controllers/api/favorite-pwa.js
juliantoledo: done.

controllers/api/favorite-pwa.js
juliantoledo: I am accepting "Authorization: eyJhbGciOiJSU..." header now.

lib/model-datastore.js
```diff
@@ -137,16 +137,37 @@ function deepCopy(object) {
...
+ * @param {number} limit
+ * @param {number} sort
+ * @return {query}
+ */
+function createQuery(kind, offset, limit, sort) {
+  const query = ds.createQuery([kind])
+    .offset(offset || 0)
+    .limit(limit);
+  if (sort) {
+    query.order(sort.field, sort.config);
```
juliantoledo: done.

juliantoledo:
@ithinkihaveacat Thank you for the review, I fix the comments. I will look deeper into using passport. Should I do it in another PR?

lib/model-datastore.js
juliantoledo: done

PR #335, opened 2017-02-17
Adds WebPerformance metrics

juliantoledo:
This new process submits a URL and sets
  pwa.webPageTest
  pwa.pageSpeed
with the corresponding info by calling an external service.

It also uses the new service to get the Lighthouse information.

We will add the new metrics to the UI in another PR.


config/config.example.json
```diff
@@ -4,7 +4,8 @@
   "CLOUD_BUCKET": "see https://console.cloud.google.com/storage/browser?project=$GCLOUD_PROJECT",
   "CLIENT_ID": "see https://console.cloud.google.com/apis/credentials?project=$GCLOUD_PROJECT",
   "CLIENT_SECRET": "see https://console.cloud.google.com/apis/credentials?project=$GCLOUD_PROJECT",
-  "LIGHTHOUSE_SERVER": "your lighthouse server URL (optional)",
+  "PAGESPEED_SERVER": "your page speed server URL (optional)",
```
ithinkihaveacat: How is the server URL used? Can we use https://www.npmjs.com/package/psi (which I think talks to https://developers.google.com/speed/docs/insights/v2/reference/pagespeedapi/runpagespeed).

models/pwa.js
```diff
@@ -101,6 +101,12 @@ class Pwa {
     return this._manifest;
   }
 
+  generateEncodedStartUrl() {
```
ithinkihaveacat: Can this just be `getEncodedStartUrl()` or similar?

models/pwa.js
```diff
@@ -101,6 +101,12 @@ class Pwa {
...
+    const parsedUrl = URL.parse(this.absoluteStartUrl);
+    this.encodedStartUrl = encodeURIComponent(parsedUrl.hostname + parsedUrl.pathname);
+    return this.encodedStartUrl;
```
ithinkihaveacat: Does this need to be cached? Maybe just regenerate every time.

config/config.example.json
juliantoledo: @ithinkihaveacat 
We have a new server that runs Lighthouse, PageSpeedInsights and WebPageTest daily, for this project and few others.

models/pwa.js
juliantoledo: We need to store as a DB indexed field it to be able to query PWAs by startURL, the idea is to use it soon to make PWA directory URLs human readable.

juliantoledo:
@andreban @sebastianbenz PTAL

PR #341, opened 2017-03-14
Initial test for rendering only the content

juliantoledo:


controllers/pwa.js
```diff
@@ -58,9 +60,14 @@ router.get('/', (req, res, next) => {
         showScore: sortOrder === 'score',
         pwaCount: pwaCount,
         startPwa: start + 1,
-        endPwa: Math.min(pwaCount, end)
+        endPwa: Math.min(pwaCount, end),
+        mainPage: true
       });
-      res.render('pwas/list.hbs', arg);
+      if (contentOnly) {
```
andreban: I think we can make this more flexible. The idea is to have different templates for the same content. 
I was playing around with this idea. Check this commit: https://github.com/andreban/gulliver/commit/3c2bd7d33ecf23bd01d9a1a8c71a447eefb89cd5

public/js/gulliver.es6.js
```diff
@@ -129,6 +129,104 @@ function setupSignedinAware() {
   }
 }
 
+function fetchInnerContent(element, newUrl) {
```
andreban: We should make this code more OO, and move it to it's own class, on a separate file, and then import from gulliver.js.

juliantoledo:
I will create another branch for these test/changes.

PR #344, opened 2017-03-17
Initial work for PRPL

juliantoledo:
- Render contentOnly templates
- Fetch only the content of the page from client


juliantoledo:
@andreban I plan to merge and keep working on another PR.

PR #347, opened 2017-03-17
Refactor Gulliver config

andreban:
Using a more Object oriented approach for Gulliver config

### DIFF
--- a/public/js/gulliver-config.js
+++ b/public/js/gulliver-config.js
@@ -1,0 +1,31 @@
+/**
+ * Copyright 2015-2016, Google, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


public/js/gulliver-config.js
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Config {
+  constructor(element) {
+    if (!element) {
+      console.log('%cConfig not found', 'color:red');
```
ithinkihaveacat: I did not know this was a thing!

public/js/gulliver-config.js
```diff
@@ -0,0 +1,31 @@
...
+      return;
+    }
+    const configJSON = JSON.parse(element.innerHTML);
+    Object.assign(this, configJSON);
```
ithinkihaveacat: Can you just inline this? The variable seems unnecessary.

public/js/gulliver-config.js
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Config {
+  constructor(element) {
+    if (!element) {
```
ithinkihaveacat: Maybe it's better to take the CSS selector here. That way if it doesn't exist the error message can report the actual selector that failed. Similarly, if the `JSON.parse(...)` fails, can report the error against the selector, instead of blowing up in an apparently unrelated location.

ithinkihaveacat:
Oh, merged already, nvm.

PR #351, opened 2017-03-27
Code refactor to prepare for Router

andreban:
- Changing styles using classes
- Refactor code for transitions


juliantoledo:
@andreban good refactor! I only found that tabs/pagination does not go offline. LGTM otherwise.

juliantoledo:
@andreban Great idea using body.setAttribute('offline', 'true') and CSS to handle offline.

public/js/offline.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Check if a Url is navigable.
+   * @param url the url to be checke for availability
+   * @returns true if the user is online or the URL is cached
+   */
+  static isAvailable(href) {
+    if (!href || window.navigator.onLine) return Promise.resolve(true);
+    return caches.match(href)
```
sebastianbenz: replace with:

```
 caches.match(href)
  .then(response => response.status === 200)
  .catch(() => false)
```

public/js/offline.js
```diff
@@ -0,0 +1,65 @@
...
+
+/* eslint-env browser */
+
+export default class Offline {
+  /**
+   * All elements with class .gulliver-online-aware will:
+   * have an 'online' dataset property that reflects the current online state.
+   * receive a 'change' event whenever the state changes.
+   */
+  static setupEventhandlers() {
```
sebastianbenz: Can we avoid static methods here? This makes it hard to test...

public/css/style.css
```diff
@@ -468,8 +473,12 @@ tr:last-child td {
   height: 56px;
   line-height: 56px;
   padding-left: 20px;
-  display: none;
   pointer-events: none;
+  transform: translateY(56px);
+  opacity: 0;
+  transition-property: transform,opacity;
```
sebastianbenz: You're using the short notation for transitions above -> use consistently

public/css/style.css
```diff
@@ -537,12 +546,32 @@ footer #github-logo {
...
   float: left;
 }
+
 .next {
   float: right;
 }
+
+/* Controls Offline and SignedIn Behavirous */
+body[offline] .gulliver-online-aware:not([cached]),
+body:not([signedIn]) .signin-online-aware-button {
```
sebastianbenz: .signin-online-aware-button => signin-aware

public/js/offline.js
```diff
@@ -0,0 +1,65 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Offline {
```
sebastianbenz: Suggestion: pass in document and window via constructor. Then you can easily mock these in tests.

public/js/offline.js
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * All elements with class .gulliver-online-aware will:
+   * have an 'online' dataset property that reflects the current online state.
+   * receive a 'change' event whenever the state changes.
+   */
+  static setupEventhandlers() {
+    const body = document.querySelector('body');
+    window.addEventListener('online', () => {
```
sebastianbenz: Move the event listener into an instance method and assign `const onlineAware = document.querySelectorAll('.gulliver-online-aware');` to a field -> no longer needs to be queried every time the status changes. Alternative is to initalize them lazy via internal getter method.

public/js/offline.js
```diff
@@ -0,0 +1,65 @@
...
+      body.removeAttribute('offline');
+      console.log('ONLINE');
+      const onlineAware = document.querySelectorAll('.gulliver-online-aware');
+      for (const e of onlineAware) {
+        e.dataset.online = JSON.stringify(true);
```
sebastianbenz: s/JSON.stringify(true)/'true'/

public/js/offline.js
```diff
@@ -0,0 +1,65 @@
...
+        e.dispatchEvent(new CustomEvent('change'));
+      }
+    });
+
+    window.addEventListener('offline', () => {
+      body.setAttribute('offline', 'true');
+      console.log('OFFLINE');
+      const onlineAware = document.querySelectorAll('.gulliver-online-aware');
+      for (const e of onlineAware) {
+        e.dataset.online = JSON.stringify(false);
```
sebastianbenz: s/JSON.stringify(false)/'false'/

public/js/offline.js
andreban: Done, thanks.

public/js/offline.js
andreban: Yes, this is waiting for refactor. I just moved the code from was inside gulliver.js to here, so it's at least in it's own code.

public/js/offline.js
andreban: This piece of code should be removed as as result from the refactors too.

public/js/offline.js
andreban: This piece of code should be removed as as result from the refactors.

public/js/offline.js
andreban: Good tip. Will do that.

public/css/style.css
andreban: Does the short notation work with multiple properties? (transform, opacity). That's the reason I used the long notation here. Looks like an alternative would be something like: 
`transition: transform 0.3s ease-in-out, opacity 0.3s ease-in-out;`, but I'm not sure it's better, as we end up repeating ourselves more.

public/css/style.css
andreban: Done. Will also rename `.gulliver-online-aware` on another PR, as it touches the code in multiple places.

public/css/style.css
sebastianbenz: Hm - good question, you could use `all` instead, not sure if that has any negative implications. Shorthand with multiple transitions works like this:

```
height 0.3s ease-out, opacity 0.3s ease 0.5s;
```

but this means you have to duplicate timing etc which is less good.

public/js/offline.js
andreban: The elements may have changed as a result from a navigation to another page. So, it's better to query the page every time. This part of the code is supposed to be removed later too.

public/css/style.css
andreban: `opacity` and `transform` are hardware accelerated. That's why I added both instead of `all`. So we avoid inadvertently animating something that is slow.

public/js/offline.js
andreban: The class is only using static methods, and most of the code will die as a result of the next refactor. So, skipping to next PR.

PR #352, opened 2017-03-28
Cache rendered pages

juliantoledo:
- Cache rendered pages of PWAs in memcache
- Cache rendered detail page for each PWA in memcache
- Invalidates cache when new PWA is added
- Invalidates cache when PWA is updated


controllers/pwa.js
```diff
@@ -163,30 +160,141 @@ router.post('/add', (req, res, next) => {
...
+        contentOnly: contentOnly
       });
+      return render(res, 'pwas/list.hbs', arg);
+    });
+}
+
+/**
+ * Cache individual PWAs from the list page if not in cache already.
+ */
+function cachePwasFromList(req, res, pwas) {
```
andreban: I think this is something we solve better from the frontend, by pre-caching links the user is looking at.

controllers/pwa.js
```diff
@@ -22,51 +22,46 @@ const lighthouseLib = require('../lib/lighthouse');
...
-  const start = parseInt(req.query.start, 10) || (pageNumber - 1) * LIST_PAGE_SIZE;
-  const limit = parseInt(req.query.limit, 10) || LIST_PAGE_SIZE;
-  const end = pageNumber * LIST_PAGE_SIZE;
-  const contentOnly = false || req.query.contentOnly;
-  let pwaCount = 0;
-  pwaLib.count()
-    .then(count => {
-      pwaCount = count;
-      return pwaLib.list(start, limit, sortOrder);
+  cache.get(req.originalUrl)
```
andreban: We may create the rendered page cache layer as a middleware. When creating the middleware, we can define which routes we want this middleware to handle. This would be a more generic approach and would leave the PWA controller to only care about rendering pages that are not cached.

controllers/pwa.js
```diff
@@ -163,30 +160,141 @@ router.post('/add', (req, res, next) => {
...
+  cache.get(url)
+    .then(cachedPage => {
+      console.log('Served from memcache: ' + url);
+      res.send(cachedPage);
+      return cachedPage;
+    })
+    .catch(_ => {
+      console.log('Not in memcache: ' + url);
+    })
+    .then(cachedPage => {
```
andreban: Looks like this is always rendering the page again, even if it's cached. We're also removing from the cache when the PWA is updated. Maybe this is not necessary.

controllers/pwa.js
juliantoledo: Yes, in the frontend for the current user. However, I think it helps to keep recent PWAs fresh for following users that hit a full server side page directly.

controllers/pwa.js
```diff
@@ -118,6 +114,7 @@ router.post('/add', (req, res, next) => {
       return pwaLib.createOrUpdatePwa(pwa);
     })
     .then(savedData => {
+      cache.flush();
```
andreban: Does this flush the whole cache? Maybe we flush only this PWA and the recently added pages.

controllers/pwa.js
juliantoledo: done.

controllers/pwa.js
juliantoledo: we should flush only the lists, I need to find if I can get the cache.keys to look for them or store them somewhere when they get cached.

controllers/pwa.js
```diff
@@ -22,51 +22,46 @@ const lighthouseLib = require('../lib/lighthouse');
 const Pwa = require('../models/pwa');
 const router = express.Router(); // eslint-disable-line new-cap
 const libMetadata = require('../lib/metadata');
+const cache = require('../lib/data-cache');
 
 const LIST_PAGE_SIZE = 32;
 const DEFAULT_PAGE_NUMBER = 1;
 const DEFAULT_SORT_ORDER = 'newest';
 
+const CACHE_LIFETIME = 60 * 60 * 24; // 24 hours
```
andreban: This looks too aggressive, specially for last updated and top score pages. We may want to have different timeouts for PWA and last updated pages.

juliantoledo: The list pages update when a user visits them (if they have changed), but I agree, I can lower it to about 6h.

views/includes/head.hbs
```diff
@@ -26,8 +26,8 @@
 <link rel="shortcut icon" href="/favicons/favicon.ico">
 <meta name="msapplication-config" content="/favicons/browserconfig.xml">
 <meta name="theme-color" content="#7cc0ff">
-<link rel="preconnect" href="https://storage.googleapis.com">
-<link rel="preconnect" href="https://apis.google.com">
+<link rel="preconnect" href="https://storage.googleapis.com" crossorigin>
```
andreban: Could this be in a separate PR?

controllers/pwa.js
```diff
@@ -195,6 +303,7 @@ router.use((err, req, res, next) => {
   // Format error and forward to generic error handler for logging and
   // responding to the request
   err.response = err.message;
+  console.log(err);
```
andreban: s/console.log/console.error

views/includes/head.hbs
juliantoledo: sure.

controllers/pwa.js
andreban: To check if the page has changed, we're rendering again and comparing before updating the cache. So, for every hit, we're loading data from the database again. This is not a problem now, but may create a performance issue if we get a traffic spike. What happens if we get to 50 requests / second?

controllers/pwa.js
andreban: There's also the problem of rendering with / without the shell. Rendering with the shell will only help users who are arriving directly to the page. Rendering without the shell will only help users navigating from the list. But since we are creating the cache on what's being rendered on the list, this shouldn't have much impact on the first case.

controllers/pwa.js
juliantoledo: done. I am storing the list in memcache and flushing only list pages.

controllers/pwa.js
juliantoledo: good point, I am not sure if it is a good idea to render both. I am deleting both on add/update.

controllers/pwa.js
juliantoledo: the logic is a bit different for both cases, let me see what I can do.

juliantoledo: @andreban I just added the middleware for caching.

controllers/pwa.js

controllers/pwa.js
juliantoledo: done.

controllers/pwa.js
juliantoledo: done. removed for now.

juliantoledo:
Fixed the comments.

PR #353, opened 2017-03-29
Rename Offline to ConnectionManager

andreban:
- Make the ConnectionManager a proper class
- Rename class `gulliver-online-aware` to `offline-aware`
- Remove code to create change events for offline

### DIFF
--- a/public/js/connectivity-manager.js
+++ b/public/js/connectivity-manager.js
@@ -15,32 +15,26 @@
 
 /* eslint-env browser */
 
-export default class Offline {
+export default class ConnectivityManager {
+
+  constructor(window) {
+    this.window = window;
+    this._setupEventhandlers();
+  }
+
   /**
    * All elements with class .gulliver-online-aware will:
    * have an 'online' dataset property that reflects the current online state.
    * receive a 'change' event whenever the state changes.
    */
-  static setupEventhandlers() {
-    const body = document.querySelector('body');
-    window.addEventListener('online', () => {
+  _setupEventhandlers() {
+    const body = this.window.document.querySelector('body');
+    this.window.addEventListener('online', () => {
       body.removeAttribute('offline');
-      console.log('ONLINE');
-      const onlineAware = document.querySelectorAll('.gulliver-online-aware');
-      for (const e of onlineAware) {
-        e.dataset.online = JSON.stringify(true);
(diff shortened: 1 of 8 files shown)


public/js/connectivity-manager.js
```diff
@@ -49,8 +43,8 @@ export default class Offline {
    * @param url the url to be checke for availability
    * @returns true if the user is online or the URL is cached
    */
-  static isAvailable(href) {
-    if (!href || window.navigator.onLine) return Promise.resolve(true);
+  isAvailable(href) {
+    if (!href || this.window.navigator.onLine) return Promise.resolve(true);
     return caches.match(href)
```
ithinkihaveacat: I think it would be better to search the `toolbox.options.cache.name` cache only here, to prevent confusing bugs in future. `caches.match(.)` searches all caches, but the SW (and `sw-toolbox`) only searches the `toolbox.options.cache.name` cache. So if there's some synchronisation problem with the caches, `isAvailable()` could return true, even though the resource will not be returned by the SW.

public/js/connectivity-manager.js
```diff
@@ -49,8 +43,8 @@ export default class Offline {
    * @param url the url to be checke for availability
    * @returns true if the user is online or the URL is cached
    */
-  static isAvailable(href) {
-    if (!href || window.navigator.onLine) return Promise.resolve(true);
+  isAvailable(href) {
```
ithinkihaveacat: Since this exists, remove the [`toolbox.router.head(.)` route](https://github.com/GoogleChrome/gulliver/blob/master/public/sw.js#L61-L80)? I don't think it's needed.

public/js/connectivity-manager.js
andreban: Hum. Not sure how to get the value of `toolbox.options.cache.name` at this point.

sebastianbenz: You could make the cache name part of your config and inject it into sw and clients.

andreban: Sounds like a plan.

public/js/connectivity-manager.js
andreban: SG. I was waiting to check if it's not used somewhere else before removing.

PR #354, opened 2017-03-30
Temporary fix for form submission

andreban:
When loading the form as a result of a transition, it needs to
dynamically load the pwa-form.js for it to work. I've implemented
a temporary and ugly solution, that needs to be replaced with proper
module loading before merging PRPL to master.


juliantoledo:
@andreban I tested the flow without issues. Sounds good, let's hammer those before merging.

PR #357, opened 2017-03-31
Refactor Analytics code

andreban:
- Create analytics.js and move analytics code there.
- Create a trackPageView method, that will be used when pages
  are changed client-side from the router.
- Change CSP header to allow importing the GA script from HTML.
- Add GA script to footer as async.

### DIFF
--- a/public/js/analytics.js
+++ b/public/js/analytics.js
@@ -16,8 +16,8 @@
 /* eslint-env browser */
 
 export default class Analytics {
-  constructor(navigator, window, config) {
-    this.navigator = navigator;
+  constructor(window, config) {
+    this.navigator = window.navigator;
     this.window = window;
     this.config = config;
     this._init();
--- a/public/js/analytics.js
+++ b/public/js/analytics.js
@@ -1,0 +1,55 @@
+/**
+ * Copyright 2015-2016, Google, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


views/includes/footer.hbs
```diff
@@ -28,3 +28,4 @@
   </a>
   </ul>
 <footer>
+<script async src="https://www.google-analytics.com/analytics.js"></script>
```
sebastianbenz: Would it make sense to create this programmatically in analytics.js to have it fully encapsulated?

public/js/analytics.js
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Analytics {
+  constructor(navigator, window, config) {
```
sebastianbenz: Why navigator as a separate parameter? You could also use window.navigator?

andreban: Good point. I totally forgot we could do `window.navigator`

views/includes/footer.hbs
andreban: I had the same thoughts, but found that most examples available include the analytics script directly on the HTML, so decided to go with this solution for now. We may experiment with moving it to analytics.js in the future.

juliantoledo: sounds good, we should find a good way to load few of the scripts we have.

PR #358, opened 2017-03-31
Tests for Cache controller

juliantoledo:
Test for Cache Controller

### DIFF
--- a/test/controllers/cache.js
+++ b/test/controllers/cache.js
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2015-2016, Google, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


lib/pwa.js
```diff
@@ -198,8 +198,12 @@ exports.createOrUpdatePwa = function(pwa) {
       }
       // Delete modified PWA from cache
       const url = '/pwas/' + pwa.id;
-      cache.del(url);
-      cache.del(url + '?contentOnly=true');
+      cache.del(url).catch(_ => {
+        // URL not in cache.
```
andreban: Should the catch handler do something, such as logging? Otherwise, we can just remove the catch block.

lib/pwa.js
```diff
@@ -198,8 +198,12 @@ exports.createOrUpdatePwa = function(pwa) {
...
+      });
+      cache.del(url + '?contentOnly=true').catch(_ => {
+        // URL not in cache.
```
andreban: See previous comment.

lib/pwa.js
juliantoledo: done, I removed both. I was trying to get rid of 2 promise unhandled warnings on the test, but this did not help.

lib/pwa.js
juliantoledo: done

juliantoledo:
@andreban I just removed the two catch blocks.

PR #361, opened 2017-04-03
Adds Router pattern to Gulliver

andreban:
- Routes controls outgoing links
- Finds proper route and coordinates the transition
- App Shell reacts to transition
- Using a simple fade-in / fade-out for transitions


public/js/routing/route.js
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Route {
+  constructor(matchRegex) {
+    this.matchRegex = matchRegex;
```
sebastianbenz: nit: mark matchRegex as internal

public/js/routing/route.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  matches(url) {
+    return this.matchRegex.test(url);
+  }
+
+  getContent(url) {
```
sebastianbenz: s/getContent/fetchContent

-> `get` indicates returning internal state, but here potentially more is happening.

public/js/routing/route.js
```diff
@@ -0,0 +1,46 @@
...
+    const contentUrl = this._getContentOnlyUrl(url);
+    return fetch(contentUrl)
+      .then(response => response.text());
+  }
+
+  transitionOut(container) {
```
sebastianbenz: I'd prefer composition here -> TransitionStrategy across routes

public/js/routing/router.js
```diff
@@ -0,0 +1,97 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Router {
+  constructor(window, shell, container) {
+    this.routes = new Set();
```
sebastianbenz: why set and not array?

public/js/routing/router.js
```diff
@@ -0,0 +1,97 @@
...
+    this.shell = shell;
+    this.window = window;
+    this.container = container;
+
+    // Update UI when back is pressed.
+    this.window.addEventListener('popstate', this._updateContent.bind(this));
+    this._control(this.window.document);
+  }
+
+  _findRoute(url) {
```
sebastianbenz: alternative:

```
routes.find(route => route.matches(url));
```

public/js/routing/router.js
```diff
@@ -0,0 +1,97 @@
...
+      // TODO: navigate to 404?
+      return;
+    }
+
+    page.transitionOut(this.container);
+    page.getContent(location)
+      .then(content => {
+        this.container.innerHTML = content;
+        this.shell.afterAttach(page);
+        this.window.scrollTo(0, 0);
```
sebastianbenz: Does this mean focus position won't be safed when going back? Would this make more sense in `transitionIn`?

public/js/routing/router.js
```diff
@@ -0,0 +1,97 @@
...
+    this.routes.add(route);
+  }
+
+  navigate(url) {
+    console.log('Navigating To: ', url);
+    this.window.history.pushState(null, null, url);
+    this._updateContent();
+  }
+
+  _control(root) {
```
sebastianbenz: maybe:

a/_control/_takeOverAnchorLinks

public/js/shell.js
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env browser */
+
+export default class Shell {
+  constructor(document) {
+    this.document = document;
+    this.backlink = document.querySelector('#backlink');
```
sebastianbenz: mark properties as internal

public/js/routing/route.js
andreban: `fetchContent` may work in this case, as we're always fetching from the network, but I feel it exposes to much of the Route behaviour, as we could be just returning static HTML, or something similar. Would something like `retrieveContent` work better?

public/js/routing/route.js
andreban: I agree 100%. I'll work on proper transition code on the next PR. Leaving this simple fadeIn / fadeOut as a placeholder.

public/js/routing/router.js
andreban: I agree. We'll need to improve this when adding proper transitions. There's also the problem of animating back. (eg: from list to detail, and back).

public/js/routing/router.js
andreban: Unfortunately, find doesn't work with Sets

public/js/routing/router.js
andreban: It doesn't make sense to add the same route twice, therefore a Set. An Array wouldn't be an issue as well, since it would just match the first occurrence of that route.

public/js/routing/route.js
andreban: Fixed.

public/js/routing/router.js
andreban: Done.

public/js/shell.js
andreban: Fixed.

public/js/routing/router.js
andreban: Done.

public/js/routing/route.js
andreban: Done.

public/js/routing/route.js
sebastianbenz: What I suggested was simply creating a new class FadeInOutTransitionStrategy and pass it in via constructor to each of the routes.

public/js/routing/router.js
sebastianbenz: Adding the same route twice is actually an user error which should result in an exception or am I missing a use case?

andreban: Actually, an Array makes more sense, as it will guarantee the order in which routes are added.

public/js/gulliver.es6.js
```diff
@@ -55,6 +59,44 @@ class Gulliver {
     this.analytics.trackPageView(window.location.href);
   }
 
+  _addRoute(regexp, shellState) {
```
sebastianbenz: nit: move _addRoute after _setupRout to reflect the call order

public/js/routing/router.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  addRoute(route) {
+    this._routes.push(route);
+  }
+
+  navigate(url) {
+    console.log('Navigating To: ', url);
+    this._window.history.pushState(null, null, url);
+    this._updateContent();
```
ithinkihaveacat: Is this necessary? Won't the `popstate` handler (the one in the constructor) pick the state change up?

public/js/gulliver.es6.js
```diff
@@ -55,6 +59,44 @@ class Gulliver {
...
+    const route = new Route(regexp);
+    this.shell.addState(route, shellState);
+    this.router.addRoute(route);
+  }
+
+  _setupRoutes() {
+    // Route for `/pwas/add`.
+    this._addRoute(/\/pwas\/add\//, {
+      showTabs: false,
```
ithinkihaveacat: Can this just be `tabs`? (Why the `show`?)

public/js/routing/route.js
andreban: Done. Added a transitions.js. I think we can keep multiple strategies there, or change later if it becomes to bukly.

public/js/routing/router.js
andreban: It is (just double checked by removing the call). Calling pushState does not make a popstate event happen. 
From [MDN: ](https://developer.mozilla.org/en-US/docs/Web/Events/popstate)

> Note that just calling history.pushState() or history.replaceState() won't trigger a popstate event. The popstate event will been triggered by doing a browser action such as a click on the back or forward button (or calling history.back() or history.forward() in JavaScript).

ithinkihaveacat: Oh, that's unfortunate, glad you checked!

juliantoledo:
@andreban great work!
I only found that the add button does not work

andreban:
@juliantoledo Fixed the "Add PWA button".

juliantoledo:
@andreban thanks, it works now.

PR #363, opened 2017-04-04
Handle clicks to open links in a new tab correctly

andreban:
- Changed router not to intercept anchor clicks when those clicks
   are generated by a button other than the left button, or if
   there's a modified on the button click. This will properly
   handle middle button clicks and Cmd+left clicks that opens
   links in a new tab.

### DIFF
--- a/public/js/routing/router.js
+++ b/public/js/routing/router.js
@@ -67,6 +67,10 @@ export default class Router {
   _takeOverAnchorLinks(root) {
     root.querySelectorAll('a').forEach(element => {
       element.addEventListener('click', e => {
+        if (e.button !== 0 || e.ctrlKey || e.metaKey || e.shiftKey || e.altKey) {
+          return true;
+        }
+
         // Link does not have an url.
         if (!e.currentTarget.href) {
           return true;
--- a/public/js/routing/router.js
+++ b/public/js/routing/router.js
@@ -64,10 +64,14 @@ export default class Router {
     this._updateContent();
   }
 
+  _isLeftClickWithoutModifiers(e) {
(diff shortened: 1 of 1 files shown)


public/js/routing/router.js
```diff
@@ -67,6 +67,10 @@ export default class Router {
   _takeOverAnchorLinks(root) {
     root.querySelectorAll('a').forEach(element => {
       element.addEventListener('click', e => {
+        if (e.button !== 0 || e.ctrlKey || e.metaKey || e.shiftKey || e.altKey) {
```
sebastianbenz: suggestion: extract into a method `isNotLeftClick(e)`

public/js/routing/router.js
```diff
@@ -64,10 +64,14 @@ export default class Router {
...
 
+  _isLeftClickWithoutModifiers(e) {
+    return e.button === 0 && !e.ctrlKey && !e.metaKey && !e.shiftKey && !e.altKey;
+  }
+
   _takeOverAnchorLinks(root) {
     root.querySelectorAll('a').forEach(element => {
       element.addEventListener('click', e => {
-        if (e.button !== 0 || e.ctrlKey || e.metaKey || e.shiftKey || e.altKey) {
+        if (!this._isLeftClickWithoutModifiers(e)) {
```
sebastianbenz: suggestion: rename to _isLeftClick**With**Modifiers - this way you don't need the `!` and the predicate becomes a lot more readable.

andreban: Changed to `_isNotLeftClickWithoutModifiers`

sebastianbenz:
Thanks!

PR #364, opened 2017-04-06
Improve variable and method names

andreban:
- Rename instances of `page` variables to `route`.
- Rename `afterUpdate` method on Shell to `updateState`.

### DIFF
--- a/public/js/shell.js
+++ b/public/js/shell.js
@@ -24,8 +24,8 @@ export default class Shell {
     this._states = new Map();
   }
 
-  addState(page, state) {
-    this._states.set(page, state);
+  addState(route, state) {
+    this._states.set(route, state);
   }
 
   _showElement(element, visible) {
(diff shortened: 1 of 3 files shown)


public/js/shell.js
```diff
@@ -24,8 +24,8 @@ export default class Shell {
     this._states = new Map();
   }
 
-  addState(page, state) {
-    this._states.set(page, state);
+  addState(route, state) {
```
sebastianbenz: suggestions: 

* rename `addState` to `setStateForRoute`
* rename `state` to `shellState`

public/js/shell.js
```diff
@@ -49,8 +49,8 @@ export default class Shell {
     tab.classList.remove('activetab');
   }
 
-  afterAttach(page) {
-    const options = this._states.get(page);
+  updateState(route) {
```
sebastianbenz: suggestion: s/updateState/onRouteChange/ -> updateState is an implementation detail, onRouteChange makes it more obvious when this should be called.

public/js/shell.js
andreban: done.

public/js/shell.js
andreban: done.

PR #365, opened 2017-04-06
Load Module dynamically with Webpack

andreban:
**** DO NOT MERGE ***

I've tried dynamically loading modules with Rollup + import syntax, but was having trouble. Decided to try out WebPack to do it, and it seems to work fine. The code is wired up to "just make it work", as I think we should discuss if it's worth replacing Rollup with WebPack.

Using rollup, gulliver.js is about 51k, and WebPack adds about 20k to the weight of gulliver.js (becoming about 73k). It's still below the [Performance Budget of 100k](https://medium.com/webpack/webpack-performance-budgets-13d4880fbf6d), And we could probably optimized other parts of the code to load dynamically too. 

There are other things we can do, such as using the `[env](https://github.com/babel/babel-preset-env)` preset for babel, so we can target more recent browsers and avoid polyfilling those.


public/js/lighthouse-chart.js
```diff
@@ -22,17 +22,41 @@ import Loader from './loader';
...
+      console.log('Googler Charts Loader already loaded');      
+      return Promise.resolve();
+    }
+
+    console.log('Loading Googler Charts Loader');
+    return new Promise((resolve, reject) => {
+      const script = document.createElement('script');
+      script.defer = true;
+      script.src = 'https://www.gstatic.com/charts/loader.js';
+      script.onload = () => {
```
ithinkihaveacat: Can be just `script.onload = resolve`.

public/js/lighthouse-chart.js
```diff
@@ -22,17 +22,41 @@ import Loader from './loader';
...
+      document.querySelector('head').appendChild(script);
+    });
+  }
+
   load() {
     this.loader.show();
-    google.charts.load('current', {packages: ['annotationchart']});
-    google.charts.setOnLoadCallback(this.drawChart.bind(this));
+    this._loadChartsApi()
+      .then(() => {
```
ithinkihaveacat: If `_loadChartsApi()` resolved to `google.charts` itself, we we could ignore the fact that that horrid global even exists:

```js
this._loadChartsAPI().then((charts) => { charts.load(...); });
```

(Could also rename it `getChartsAPI()`.)

public/js/gulliver.es6.js
```diff
@@ -60,8 +56,8 @@ class Gulliver {
     this.analytics.trackPageView(window.location.href);
   }
 
-  _addRoute(regexp, transitionStrategy, shellState) {
-    const route = new Route(regexp, transitionStrategy);
+  _addRoute(regexp, transitionStrategy, shellState, onAttached) {
```
ithinkihaveacat: Maybe add docs for this method? It's getting big.

andreban: This definitely needs refactor. But I wanted to discuss the whole WebPack, dynamic loading thing first.

PR #366, opened 2017-04-07
Inlines chart.js and form.js in gulliver

andreban:
Adds `lighthouse-chart.js` and `pwa-form.js` as static imports to Gulliver so that the chart and the form work properly when being rendered from the client-side.

Adding both scripts to `gulliver.js` makes it grow by only 2.7k.

### DIFF
--- a/public/js/gulliver.es6.js
+++ b/public/js/gulliver.es6.js
@@ -27,7 +27,6 @@ import 'yaku/dist/yaku.browser.global.min.js';
 // https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/fetch/config.json
 import 'whatwg-fetch/fetch';
 
-import './loader.js';
 import Messaging from './messaging';
 import NotificationCheckbox from './ui/notification-checkbox';
 import Config from './gulliver-config';
@@ -39,6 +38,8 @@ import Router from './routing/router';
 import Route from './routing/route';
 import Shell from './shell';
 import FadeInOutTransitionStrategy from './routing/transitions';
+import PwaForm from './pwa-form';
+import LighthouseChart from './lighthouse-chart';
 
 class Gulliver {
   constructor() {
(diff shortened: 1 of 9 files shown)


public/js/gulliver.es6.js
```diff
@@ -55,43 +56,57 @@ class Gulliver {
...
-    this._addRoute(/\/pwas\/(\d+)/, fadeInOutTransitionStrategy, {
+    this._addRoute(/\/pwas\/(\d+)/, fadeInOutTransitionStrategy, setupLighthouseChart, {
       showTabs: false,
       backlink: true,
       subtitle: false
     });
 
     // Route for `/?sort=score`.
-    this._addRoute(/\/\?.*sort=score/, fadeInOutTransitionStrategy, {
+    this._addRoute(/\/\?.*sort=score/, fadeInOutTransitionStrategy, null, {
```
sebastianbenz: suggestion, create a dummy function called `nop` and pass it instead of null. This makes it more obvious that there is nothing to do.

public/js/lighthouse-chart.js
```diff
@@ -22,17 +22,39 @@ import Loader from './loader';
...
+    return new Promise((resolve, reject) => {
+      const script = document.createElement('script');
+      script.defer = true;
+      script.src = 'https://www.gstatic.com/charts/loader.js';
+      script.onload = () => {
+        console.log(window.google);
+        resolve(window.google);
+      }
+      script.onerror = reject;
+      document.querySelector('head').appendChild(script);
```
sebastianbenz: s/document.querySelector('head')/document.head/

public/js/gulliver.es6.js
andreban: Done.

public/js/lighthouse-chart.js
andreban: Done.

PR #370, opened 2017-04-19
Serve the Empty Shell from the Service Worker

andreban:
- Serves the Empty Shell from the Service Worker.
- Loads the Shell content from the client side.

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -60,7 +60,7 @@ toolbox.router.get(/.*\.(js|png|svg|jpg|css)$/, (request, values, options) => {
 });
 
 /**
- * Utility method to retrueve a url from the `toolbox.options.cache.name` cache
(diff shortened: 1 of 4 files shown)


public/sw.js
```diff
@@ -57,21 +59,41 @@ toolbox.router.get(/.*\.(js|png|svg|jpg|css)$/, (request, values, options) => {
   return toolbox.cacheFirst(request, values, options);
 });
 
-toolbox.router.default = (request, values, options) => {
+/**
+ * Utility method to retrueve a url from the `toolbox.options.cache.name` cache
+ *
+ * @param {*} url url to be requested fromt he cache.
```
juliantoledo: s/retrueve/retrieve

andreban: Fixed.

PR #375, opened 2017-04-26
Add new Charts to detail page

juliantoledo:


public/js/charts.js
```diff
@@ -62,7 +61,7 @@ export default class LighthouseChart {
       return;
     }
     const pagewith = Math.max(document.documentElement.clientWidth, window.innerWidth || 0);
-    fetch(CHART_BASE_URL + pwaId)
+    fetch(this.baseUrl + pwaId + '?graph=true')
```
andreban: The chart should receive the whole URL, and not have to worry about assembling it. 

If we want to be more generic,  we could pass a datasource instead of a Url. 

```
class UrlDatasource {
  contructor(url) {
    this._url = url;
  }
  
  get() {
    return fetch(this._url);
  }
}
```
and then do:

```
datasource.get().then(...)
```
I'd imagine something like this if we decide to pass the data inside a `script type=application/json` section to avoid another request. Might be an overkill tough

public/js/gulliver.es6.js
```diff
@@ -85,12 +85,14 @@ class Gulliver {
       subtitle: true
     });
 
-    const setupLighthouseChart = () => {
-      new LighthouseChart().load();
+    const setupCharts = () => {
+      new Charts('lighthouse-chart', '/api/lighthouse/graph/').load();
+      new Charts('psi-chart', 'https://web-performance-dot-pwa-directory.appspot.com/pagespeedreport/').load();
+      new Charts('wpt-chart', 'https://web-performance-dot-pwa-directory.appspot.com/webpagetestreport/').load();
```
sebastianbenz: Suggestion: change this to

```
const chartConfigurations = [
  {id: 'lighthouse-chart', url: '/api/lighthouse/graph/'},
  {id: 'psi-chart', url: 'https://web-performance-dot-pwa-directory.appspot.com/pagespeedreport/'},
  {id: 'wpt-chart', url: 'https://web-performance-dot-pwa-directory.appspot.com/webpagetestreport/'},
];

chartConfigurations.forEach(config => new Chart(config).load())
```

This makes it easier to change the chart API and doesn't repeat the initialization logic. This requires also changing the chart constructor to accept a config object.

public/js/gulliver.es6.js
```diff
@@ -39,7 +39,7 @@ import Route from './routing/route';
 import Shell from './shell';
 import FadeInOutTransitionStrategy from './routing/transitions';
 import PwaForm from './pwa-form';
-import LighthouseChart from './lighthouse-chart';
+import Charts from './charts';
```
sebastianbenz: s/Charts/Chart

public/js/charts.js
juliantoledo: Good point.

public/js/gulliver.es6.js
juliantoledo: done.

public/js/gulliver.es6.js
juliantoledo: done.

public/js/charts.js
juliantoledo: I fixed the URL generation, but I think the datasource is an overkill for now.

juliantoledo:
Fixed comments.

@sebastianbenz I think is a good idea to merge the templates.

juliantoledo:
Fixed the "Google Charts loader.js can only be loaded once" error

public/js/chart.js
```diff
@@ -20,49 +20,51 @@ import Loader from './loader';
...
       }
-      script.onerror = reject;
-      document.head.appendChild(script);
     });
   }
+
   load() {
     this.loader.show();
-    this._loadChartsApi()
-      .then(google => {
```
andreban: We were using resolve(window.google) on the _loadChartsApi promise previously. The reason was that by doing this we could avoid accessing the global on the window object from here. I know you've been through some issues making this work, but wondering if we could keep that.

juliantoledo: @andreban done

PR #376, opened 2017-04-26
Adds client js test for event-target

### LINKED ISSUE #372: Add code testing structure for the frontend code (public/js)
Code in `public/js` is not playing well with our testing configuration, as it uses the ES6 import syntax and we get something like following:

```
gulliver/test/js/event-target.js:18
import EventTarget from '../../public/js/event-target';
^^^^^^
SyntaxError: Unexpected token import
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at /usr/local/lib/node_modules/mocha/lib/mocha.js:222:27
    at Array.forEach (native)
    at Mocha.loadFiles (/usr/local/lib/node_modules/mocha/lib/mocha.js:219:14)
    at Mocha.run (/usr/local/lib/node_modules/mocha/lib/mocha.js:487:10)
    at Object.<anonymous> (/usr/local/lib/node_modules/mocha/bin/_mocha:459:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149
```

andreban:
- Split tests between `app` and `client`.
- Updates .babelrc to support `default` and `test` environments.
- Update package.json
  - Adds 'mocha-client` target to test the client js.
  - Renames `mocha` to `mocha-app`.
  - `test` target invokes both `mocha-client` and `mocha-app`.

This will resolve #372


test/client/js/event-target.js
```diff
@@ -0,0 +1,126 @@
...
+
+/* global describe it */
+'use strict';
+import EventTarget from '../../../public/js/event-target';
+const assert = require('assert');
+
+describe('js.event-target', () => {
+  describe('#addEventLister', () => {
+    it('Add a single listener to a type', () => {
+      const eventTarget = new EventTarget();
```
sebastianbenz: extract the object creation into a before block

test/client/js/event-target.js
```diff
@@ -0,0 +1,126 @@
...
+      const callback = () => {};
```
sebastianbenz: move callback declaration into before block

test/client/js/event-target.js
```diff
@@ -0,0 +1,126 @@
...
+      eventTarget.addEventListener('type', callback);
+      const eventListeners = eventTarget.getEventListeners('type');
+      assert.notEqual(eventListeners, null);
```
sebastianbenz: this assert Is not needed

test/client/js/event-target.js
```diff
@@ -0,0 +1,126 @@
...
+      assert.equal(eventListeners.size, 1);
+      assert.equal(eventListeners.has(callback), true);
```
sebastianbenz: I wouldn't test whether an callback exists (or whether event listeners has size 1. These are implementation details. Testing implementation details make tests fragile. Instead focus tests on the behavior that is visible from the outside -> event dispatching. I

andreban:
Rewrote the tests.

test/client/js/event-target.js
```diff
@@ -0,0 +1,53 @@
...
+
+  afterEach(() => {
+    simpleMock.restore();
+    callbackA.reset();
+  });
+
+  it('Fires the correct callback', done => {
+    eventTarget.addEventListener('event-a', callbackA);
+    eventTarget.dispatchEvent({type: 'event-a'});
+    setTimeout(() => {
```
sebastianbenz: I don't think we need a timeout here.

ithinkihaveacat: Looking at https://mochajs.org/#asynchronous-code and also https://lostechies.com/derickbailey/2012/08/17/asynchronous-unit-tests-with-mocha-promises-and-winjs/ I think the right way to do this is to pass `done` to the callback, and arrange for it to be called.

The problem with removing `setTimeout()` completely is that `dispatchEvent()` might (for some reason) only process events after a timeout, in which case the `done()` has already happened. Mocha apparently imposes its own 2 sec timeout for `done()` to happen, so I think this test should look like:

```js
it("Fires the correct callback 2", done => {
  const expectedEvent = { type: "event-a" };
  eventTarget.addEventListener("event-a", actualEvent => {
    assert.deepEqual(actualEvent, expectedEvent);
    done();
  });
  eventTarget.dispatchEvent(expectedEvent);
});
```

(`before()` should probably also be changed to `beforeEach()` to get a fresh `eventTarget` each time, as I discovered when adding a "Fires the correct callback 2" test...)

sebastianbenz: How do you check if an event hasn't been fired in this case?

Why not simply go with:

```
it("Fires the correct callback 2", () => {
  eventTarget.addEventListener('event-a', callbackA);
  eventTarget.dispatchEvent({type: 'event-a'});
  assert.equal(callbackA.callCount, 1);
});
```

Guarding against a future change in eventTarget is difficult as we don't know how it's going to change.

ithinkihaveacat: @andreban pointed that out as well. `done()` with an argument is considered an error, so that could possibly be used (instead of making the spy smarter). Getting rid of that `setTimeout` is the most important thing.

andreban:
Removed the timeouts. Fixed using beforeEach instead of before. Keeping the spy, as it makes the test clearer. Also, avoids issues with things such as the callback being invoked twice, which would fail the test with callcount = 2, but would pass when calling done on the first callback.

PR #378, opened 2017-05-04
Escape HTML from Lighthouse descriptions

juliantoledo:


juliantoledo:
@andreban sure, done.

PR #379, opened 2017-05-04
Use a Loader for the page transition

andreban:
- The fade-in / fade-out transition feels slow. Use a Loader in
  the transition to improve the effect.

### DIFF
--- a/public/js/routing/transitions.js
+++ b/public/js/routing/transitions.js
@@ -14,8 +14,9 @@
  */
 
 /* eslint-env browser */
+import Loader from '../loader';
 
-export default class FadeInOutTransitionStrategy {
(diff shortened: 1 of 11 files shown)


public/js/routing/transitions.js
```diff
@@ -26,3 +27,23 @@ export default class FadeInOutTransitionStrategy {
...
+export class LoaderTransitionStrategy {
+  constructor(window) {
+    this._window = window;
+    const loaderDiv = window.document.querySelector('.page-loader');
+    this._loader = new Loader(loaderDiv);
+  }
+
+  transitionIn(container) {
+    console.log('Transition In');
+    container.classList.remove('transition');
```
sebastianbenz: Can you extract the constant into a string

PR #380, opened 2017-05-04
Fixes offline inconsistencies

andreban:
- Properly marks links as cached after client-side load
- Marks body as offline on first load


public/js/gulliver.es6.js
```diff
@@ -68,6 +68,7 @@ class Gulliver {
     this.router.addEventListener('navigate', e => {
       this.analytics.trackPageView(e.detail.url);
       this.shell.onRouteChange(e.detail.route);
+      this.offlineSupport.markAsCached(document.querySelectorAll('.offline-aware'));
```
sebastianbenz: Can we move the string into a constant to make it easier to change the value in the future.

PR #383, opened 2017-05-04
Move compress from index to App level

juliantoledo:
We were not compressing at App level and missing the static file compression.


andreban:
LGTM

PR #386, opened 2017-05-05
Execute more than 1 tasks per cron job run

juliantoledo:
At the moment we can only execute 1 task per minute (minimum time for cron jobs), however our webperformance server can handle more.

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -20,6 +20,7 @@ const pwaLib = require('../lib/pwa');
 const tasksLib = require('../lib/tasks');
 const Task = require('../models/task');
 const router = express.Router(); // eslint-disable-line new-cap
+const promiseSequential = require('../lib/promise-sequential');
 
 const APP_ENGINE_CRON = 'X-Appengine-Cron';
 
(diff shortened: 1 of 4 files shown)


controllers/tasks.js
```diff
@@ -51,24 +52,29 @@ router.get('/cron', (req, res, next) => {
...
-        if (task) {
-          tasksLib.executePwaTask(task);
-        }
-        res.sendStatus(200);
-      })
-      .catch(err => {
-        next(err);
-      });
+    try {
+      for (var i = 0; i < tasksToExecute; i++) {
```
andreban: Use `let` instead of `var`.

controllers/tasks.js
```diff
@@ -51,24 +52,29 @@ router.get('/cron', (req, res, next) => {
...
+        tasksList.push(tasksLib.popExecute);
+      }
+      // Execute sequentially the 1 OR req.query.tasks of tasks
+      promiseSequential.all(tasksList)
```
andreban: What happens if a single task fails? Do they all fail?

controllers/tasks.js
juliantoledo: my bad, done.

controllers/tasks.js
juliantoledo: @andreban it will successfully run all until one fails, it will return an error for the cron job and log errors.

Each task has a retry mechanism, so they will be queued again, the ones after the one that fails will not be pop out of the queue, so the next cron will pick them up.

PR #400, opened 2017-05-11
Add search to the backend

juliantoledo:


controllers/cache.js
```diff
@@ -28,6 +28,12 @@ const CACHE_LIFETIME = 60 * 60 * 6; // 6 hours
  * overrides res.send to be able to cache rendered HTML before sending.
  */
 router.get('*', (req, res, next) => {
+  // do not cache search queries
+  if (req.query.search) {
```
andreban: Is the result likely to search for searches that have the same query? Otherwise, maybe we can cache the search results as well.

lib/pwa.js
```diff
@@ -82,6 +83,53 @@ exports.list = function(skip, limit, sort) {
 };
 
 /**
+ * Add all PWAs from the DB into the text search index.
+ */
+exports.indexAllPwas = _ => {
```
andreban: I'm wondering if we could extract this method go a PwaIndex class, and then just use it from here, as composition? we could also make the implementation more generic, but we might wait until it's needed.

lib/pwa.js
```diff
@@ -82,6 +83,53 @@ exports.list = function(skip, limit, sort) {
...
+  return indexPage(0, 100);
+};
+
+/**
+ * Search for PWAS using the text search index.
+ *
+ * @param {string} query
+ * @return {resultPage} resultPage with an arrays of PWAs and hasMore boolean
+ */
+exports.searchPwas = string => {
```
andreban: Same thing as previous comment.

lib/pwa.js
```diff
@@ -191,36 +239,48 @@ exports.createOrUpdatePwa = function(pwa) {
...
   ]);
 };
 
 /**
+ * Update PWA in the search index.
+ *
+ * @param {Pwa} pwa to update
+ * @return {Promise<Pwa>}
+ */
+exports.updateSearchIndex = function(pwa) {
```
andreban: Same thing.

lib/search.js
```diff
@@ -0,0 +1,166 @@
...
+  addPwa(pwa) {
+    const doc = this._docFromPwa(pwa);
+    this._index.addDoc(doc);
+    this.sarchIndexChange();
+    return Promise.resolve(doc);
+  }
+
+  addPwas(pwas) {
+    pwas.forEach(pwa => {
+      const doc = this._docFromPwa(pwa);
```
andreban: The next two lines are repeated from the addPwa method. Maybe extracting to another method and using it in both methods?

lib/pwa.js
juliantoledo: It's a good idea, I think I rather leave it here for now, it's very related to PWA.

lib/pwa.js
juliantoledo: same here.

lib/search.js
juliantoledo: done

lib/pwa.js
juliantoledo: you are right, Pwa is too big already, I am creating the new class.

lib/search.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Add a list of PWA to the search index.
+   *
+   * @param {PWA[]} PWAs to index
+   * @return {Promise<>}
+   */
+  addPwas(pwas) {
+    pwas.forEach(pwa => {
```
andreban: Can be written as:

`pwas.forEach(pwa => this.addPwa(pwa));`

controllers/cache.js
andreban: Maybe we could add a short cache for those later. The requests should be cached by the CDN anyway.

lib/search.js
juliantoledo: done

controllers/cache.js
juliantoledo: Actually I just realized that we invalidate list pages (including ?search ones) when a new PWA is added/updated. So, we can cache them.

PR #411, opened 2017-05-30
Search box

juliantoledo:
Add search to the UI


controllers/pwa.js
```diff
@@ -187,13 +187,28 @@ function renderPwaListPage(req, res) {
   const limit = parseInt(req.query.limit, 10) || LIST_PAGE_SIZE;
   const end = pageNumber * LIST_PAGE_SIZE;
   const contentOnly = false || req.query.contentOnly;
-  const search = false || req.query.search;
+
+  let mainPage = false;
```
andreban: Maybe we could group those parameters in an object and create a method to extract the data from the request and return the object.

controllers/pwa.js
```diff
@@ -187,13 +187,28 @@ function renderPwaListPage(req, res) {
...
+  let search = false;
+  let searchValue;
+  let backlink = false;
+  if (typeof req.query.search === 'undefined') {
```
andreban: Maybe we can use `/pwa/search?query=abc`, and differentiate this on the path level instead of parameter.

public/js/gulliver.es6.js
```diff
@@ -89,10 +90,13 @@ class Gulliver {
...
       showTabs: false,
       backlink: true,
-      subtitle: true
+      subtitle: true,
+      search: false
     });
 
     const setupCharts = () => {
+      // Clear search-input value
+      document.querySelector('#search-input').value = '';
```
andreban: Is there a special reason why we are setting the value of `#search-input` inside the `setupCharts` method?

public/js/gulliver.es6.js
```diff
@@ -107,24 +111,44 @@ class Gulliver {
     this._addRoute(/\/pwas\/(\d+)/, transitionStrategy, setupCharts, {
       showTabs: false,
       backlink: true,
-      subtitle: false
+      subtitle: true,
+      search: false
     });
 
-    const nop = () => {};
+    const nop = () => {
```
andreban: This is not a `nop` anymore. Should change the method name to describe what it does.

public/js/gulliver.es6.js
```diff
@@ -164,6 +188,33 @@ class Gulliver {
       window.history.back();
     });
   }
+
+ /**
+  * Setup/configure search button
+  */
+  setupSearchElements() {
```
andreban: Sounds like we could create a `search-button.js`, encapsulate this code there and just instantiate it from `gulliver.js`

public/js/gulliver.es6.js
juliantoledo: fixed. no, it can be in the router definition.

public/js/gulliver.es6.js
juliantoledo: fixed

controllers/pwa.js

public/js/gulliver.es6.js
juliantoledo: done

controllers/pwa.js

juliantoledo:
@andreban Thank you. Comments fixed.

Issue #224, opened 2016-10-21
Human readable PWA name in URL path


sebastianbenz:
For example, instead of: 

https://pwa-directory.appspot.com/pwas/5692592335355904

use the slugified version of the PWA name, e.g.:

https://pwa-directory.appspot.com/pwas/amp-by-example

juliantoledo:
- [ ] Create a slugified field for the PWA object, update all PWAs
- [ ] Create a new router to handle slugified URLs
- [ ] 301 redirect from ID URLs to slugified URLs
- [ ] Change the home screen to use slugified URLs

ithinkihaveacat:
Does https://www.npmjs.com/package/slug guarantee that unique input strings map to different "slugified" strings? We might need URLs with both the slug and the unique id, like <https://pwa-directory.appspot.com/pwas/5692592335355904/amp-by-example>.

sebastianbenz:
Maybe use a URI encoded version of the PWA URL:

```
> encodeURI('ampbyexample.com')
'ampbyexample.com'
```

These should be unique.

juliantoledo:
I am currently checking if the slugified name is unique when is generated and fallback in this order:  Name -> Description -> Id.

sebastianbenz:
That sounds a lot more complicated than simply using the URL...

juliantoledo:
@sebastianbenz very true. Let me try with the URL removing parameters.

juliantoledo:
For www.domain.com/folder/
What do you think is better?
www.domain.com/folder/
www.domain.com%2Ffolder%2F
www.domain.com-folder

sebastianbenz:
I'd go with `www.domain.com%2Ffolder%2F`:

* `www.domain.com/folder/` makes it impossible to use the path for PWA actions, e.g. `.../pwas/ampbyexample.com/edit/`
* `www.domain.com-folder` is not unique as it is also a valid URL

=> ` www.domain.com%2Ffolder%2F`

juliantoledo:
@sebastianbenz @ithinkihaveacat done, I updated #242

juliantoledo:
I created a branch to keep working on this:
https://github.com/GoogleChrome/gulliver/tree/encoded-start-url

The problem with the start_url is that it may change, for example, they could remove the folder. Once we start sharing/exporting those URLs we may need to keep track of those changes to handle redirects or create a slugifed version that never changes.

PR #417, opened 2017-06-13
use encodedUrl instead of ID for nav

samthor:
Super naïve approach to #224. This changes all URLs to the encodedUrl version and redirects from numeric IDs.

(There also seems to be no nav tests, that's fine, just couldn't update anything)

Downsides-
- slower than ID lookup (always a query)
- top-level domains all end with %2F, which is a bit ugly: e.g., "santatracker.google.com/" => "santatracker.google.com%2F"

### DIFF
--- a/controllers/pwa.js
+++ b/controllers/pwa.js
@@ -138,7 +138,7 @@ router.post('/add', (req, res, next) => {
  * Display a PWA or redirects to the encodedStartUrl of the PWA.
  */
 router.get('/:pwa', (req, res, next) => {
-  if (isNaN(+req.params.pwa)) {
+  if (isNaN(Number(req.params.pwa))) {
     // This URL is not a number, assume encodedStartUrl.
     renderOnePwa(req, res)
       .then(html => {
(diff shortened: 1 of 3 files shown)


controllers/pwa.js
```diff
@@ -135,27 +135,38 @@ router.post('/add', (req, res, next) => {
...
+    // This URL is not a number, assume encodedStartUrl.
+    renderOnePwa(req, res)
+      .then(html => {
+        res.send(html);
+      })
+      .catch(err => {
+        err.status = 404;
+        return next(err);
+      });
+  } else {
```
andreban: nit: Just a matter of code style, but we could do an early return here.

samthor: Done

andreban:
It seems `res.redirect` [defaults to a 302](https://expressjs.com/en/4x/api.html#res.redirect) (temporary). Would a 301 (permanent) would be more appropriate in this case? 

Doing `res.redirect(301, req.baseUrl + '/' + pwa.encodedStartUrl);` would to the trick.

samthor:
Sounds good, I've included it now.

andreban:
All looks good. Do you have permission to merge?

PR #427, opened 2017-06-21
fallback to previousEncodedStartUrl array for old encoded URLs

samthor:


juliantoledo:
@samthor, most of the PWAs on the list work, but I found that http://localhost:8080/pwas/www.fidgetspin.xyz%2F fails to load.

In the meeting yesterday we discussed that maybe keeping the ID in the URL could help make the whole process less error prone and it would follow a pattern many sites use, for example:
http://localhost:8080/pwas/5649231069052928/vitaliy-bobrov.github.io%2Findex.html
What do you think?

samthor:
so that example doesn't look like it's on staging—I'm not testing using prod. Is there a chance `encodedStartUrl` isn't set for all prod sites, was its creation a recent change?

It's up to your team what you think is important. We could then redirect anything under the ID to the "friendly" name, and/or still support `/pwas/encodedStartUrl` redirect to => `/pwas/id/encodedStartUrl`. WDYT? Do you still think the `previousEncodedStartUrl` is useful then?

juliantoledo:
I will check in the DB to find out if `encodedStartUrl` is set properly for that PWA and others.

The idea that came up was to use `/pwas/id/encodedStartUrl` and just ignore `encodedStartUrl` without any redirections. Do you think that would be friendly enough?

samthor:
I don't have a strong opinion. But to clarify: if I type `/pwas/<id>/randomString`, what happens?

andreban:
When we get a `/pwas/<id>/<pretty_path>`. Assuming that `<id>` is a valid one, we should fetch the PWA from db, check if `<pretty_path>` matches the expected one for that PWA and, if it doesn't, we respond with a 301 to the correct address. Otherwise we serve the page. Does this make sense?

PR #433, opened 2017-08-03
Lighthouse v2.3.0 and misc

cedricbellet:
- Switch to 2.3.0
- Chrome panick: kill the instance if Chrome is unreachable
- Misc improvements


juliantoledo:
Great changes, I think we should document (in code and/or readme) how/when the instances are killed.

I will test it locally asap.

PR #438, opened 2017-09-08
Adding readiness_checks

cedricbellet:
Added readiness checks which allow for much better control of the status of the different containers (load balancer will not send new requests to busy instances)

https://cloud.google.com/appengine/docs/flexible/java/configuring-your-app-with-app-yaml#updated_health_checks

### DIFF
--- a/lighthouse_machine/app.yaml
+++ b/lighthouse_machine/app.yaml
@@ -30,9 +30,18 @@ handlers:
   script: IGNORED
   secure: always
 
-health_check:
-  enable_health_check: True
-  check_interval_sec: 10
-  timeout_sec: 10
-  unhealthy_threshold: 2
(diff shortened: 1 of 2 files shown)


juliantoledo:
Good PR, thank you @cedricbellet, I will run few tests before approving it.

cedricbellet:
Apparently the path argument does not work as described in the documentation. I deployed to my own projects and the paths for health checks was still the default liveness_check and readiness_check endpoints. I had to manually override the paths. That being said, moving to updated health checks is still the right thing to do.

cedricbellet:
Disregard previous comment - although the path shown in the UI does not match the YAML file, it seems to be working. Manually overwriting the paths messed everything up :) I think the commit works well as is (doing some more experiments)

PR #439, opened 2017-09-20
Switch to LH v2.4.0

juliantoledo:
Switch to LH v2.4.0


juliantoledo:
@cedricbellet I am testing LH 2.4 already in my servers

PR #440, opened 2017-09-20
Task for updating unscored PWAs more frequently

juliantoledo:


controllers/tasks.js
```diff
@@ -52,6 +52,36 @@ router.get('/cron', (req, res, next) => {
 });
 
 /**
+ * GET /tasks/updateunscored
+ *
+ * We use a GET from the cron job to launch a PWA update process
+ * for all PWAs.
+ */
+router.get('/updateunscored', (req, res, next) => {
```
andreban: This looks really similar to the original cron task. Wondering if we can improve co-reuse between both.

juliantoledo: @andreban yes, I will do some refactoring.

juliantoledo:
@andreban fixed comment.

PR #451, opened 2017-11-03
Fix count.data error

juliantoledo:


juliantoledo:
Yes, it was the datastore one.

Issue #463, opened 2018-02-22
feat(Add in RSS feed of latest sites)


PaulKinlan:
I have ideas and plans. Would love to get an RSS feed of the latest.

juliantoledo:
It sounds interesting, let's see if we can do it soon.

juliantoledo:
@PaulKinlan https://pwa-directory.appspot.com/api/pwa/?format=rss&limit=100

PaulKinlan:
curl https://pwa-directory.appspot.com/api/pwa/?format=rss&limit=100 works great thanks, one thing I noticed though is that if I view it through the web, the appshell hosts the page


![screen shot 2018-03-01 at 7 21 08 am](https://user-images.githubusercontent.com/45510/36811325-7970df22-1d21-11e8-8a28-f0e2a9142b8d.png)

andreban:
I noticed it too. Created #466 sooner today to track it.

juliantoledo:
We fixed the app shell issue on #467 it should work once the SW updates.

@PaulKinlan does the way we set the content in <content:encoded> work for you? Please let us know if we should set it up differently.

PR #464, opened 2018-02-22
Adds a RSS ouput for the feed

andreban:
This adds an RSS feed, as requested on #463. There are 2 things we need to look: 
- Do we need an API_KEY for the feed? 
- Also, it seems we always deliver the entire list of PWAs. Maybe we could add parameters to paginate?

### DIFF
--- a/controllers/api/pwa.js
+++ b/controllers/api/pwa.js
@@ -20,6 +20,7 @@ require('express-csv');
 const pwaLib = require('../../lib/pwa');
 const router = express.Router(); // eslint-disable-line new-cap
 const CACHE_CONTROL_EXPIRES = 60 * 60 * 1; // 1 hour
+const RSS = require('rss');
 
 const config = require('../../config/config');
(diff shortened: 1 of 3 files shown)


juliantoledo:
Very fast PR!
- I think we should not ask for the API_KEY for the RSS.
- Yes, pagination could be good as optional and it should be sorted by 'newest'

Thank you.

PaulKinlan:
This is looking awesome.  One extra thing, we should add in a <link rel=alternate> on main page that points directly to the RSS feed. This will help people add it to their feed readers and hopefully help increase discovery. Selfishly, I'm about to deploy a site that would link to this feed :)

PaulKinlan:
Maybe it's not for this PR,

But I would love us to flesh out the content/description for each PWA, so that it includes a link to Gulliver, a Link to the start url so that they can try it out, and also include a summary of all that we know (manifest name, description, icon, speedscores,lighthouse etc.)

juliantoledo:
I will continue in another PR.

PR #472, opened 2018-03-19
RSS feed and API changes

juliantoledo:


juliantoledo:
@andreban where is the comment? I cannot find it.

controllers/api/pwa.js
```diff
@@ -128,34 +160,48 @@ const rssWriter = new RssWriter();
...
-        }
-        case 'rss': {
-          rssWriter.write(res, result.pwas);
-          break;
-        }
-        default: {
-          jsonWriter.write(res, result.pwas);
-        }
+
+  return new Promise((resolve, reject) => {
```
andreban: We could simplify this with something like:

```js
let promise;
if (req.params.id) {
  promise = findPwa(req.params.id);
} else {
  promise = pwaLib.list(...)
}
promise
  .then(result) => {
    ...
  }

```

andreban:
Failed to save it, sorry. Re-added.

controllers/api/pwa.js
juliantoledo: Good catch! fixed.

Issue #477, opened 2018-04-20
Add tab for "Installable" PWAs and make it default


andreban:
- The installability can be checked as part of [this](https://github.com/GoogleChrome/lighthouse/blob/06a0a45b478565e6fa90636a41d440a41a40b74c/lighthouse-core/audits/webapp-install-banner.js#L99
) audit.
- The results should be sorted by score and date of addition.

juliantoledo:
I guess that would be: 
![screen shot 2018-04-20 at 3 33 42 pm](https://user-images.githubusercontent.com/1920764/39053820-3f59c93c-44b0-11e8-9e2a-36494868d841.png)

andreban:
@slightlyoff, FYI

PR #478, opened 2018-04-25
[WIP]Adds an 'Installable' tab

andreban:
- Currently, using lighthouseScore >= 90 as a proxy for
  installability
- Moving tabs to paths instead of query string (eg: /?sort=newest
  to /pwas/newest)

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -36,6 +36,11 @@ router.get('/', (req, res) => {
   router.handle(req, res);
 });
 
+router.get('/installable', (req, res) => {
+  req.url = '/pwas/installable';
+  router.handle(req, res);
+});
+
 // ServiceWorker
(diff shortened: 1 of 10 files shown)


andreban:
Still needs work on the UI. Adding that to the next commit.

andreban:
Related to #477

juliantoledo:
LGTM, just one question.
Should Installable be the default? Currently http://localhost:8080/ points to newest.

andreban:
Yes, the plan is to make them the default. We can deploy the 1st version with the installable tab hidden, so we can make sure that the lighthouseScore >= 90 makes sense. Once we have that, we make it the default. WDYT?

juliantoledo:
SGTM

I can surface the installable flag from the LH reports.

PR #484, opened 2018-05-16
Uses the `installable` attribute for the Installable tab

andreban:
- Attribute is extracted from the Lighthouse report


andreban:
Related to #477

PR #501, opened 2018-06-29
Adds API endpoint to add a PWA

andreban:
- Endpoint to be used for async upload form the frontend.
- Does not support adding from 3P domains
- Expects a JSON with the data.

### DIFF
--- a/controllers/api/pwa.js
+++ b/controllers/api/pwa.js
@@ -23,8 +23,10 @@ const router = express.Router(); // eslint-disable-line new-cap
 const verifyIdToken = require('../../lib/verify-id-token');
 const bodyParser = require('body-parser');
 const Pwa = require('../../models/pwa');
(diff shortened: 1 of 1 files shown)


controllers/api/pwa.js
```diff
@@ -193,4 +196,33 @@ router.get('/:id*?', (req, res) => {
...
+router.post('/add', bodyParser.json(), (req, res) => {
+  const idToken = req.body.idToken;
+
+  if (!idToken) {
+    res.sendStatus(401, JSON.stringify({error: 'user not logged in'}));
+    return;
+  }
+
+  const manifestUrl = req.body.manifestUrl;
+  if (!manifestUrl) {
```
juliantoledo: I think it would be also good to check that this is a proper URL.

andreban: updated.

PR #503, opened 2018-07-06
Changes `/pwa/add` to use AJAX

andreban:
- Adding a PWA can take a while (eg; fetch and parse manifest
  from server).
- This makes the post between pages feel like the app is broken.
- Moved to use AJAX and a spinner so that the app doesn't feel
  broken.
- Also allows adding another PWA while the first is being uploaded.

Current version deployed here: https://gulliver-1371.appspot.com/pwas/add


andreban:
@fstanis, FYI

Issue #500, opened 2018-06-28
Create a Share Target


PaulKinlan:
I find pwa's as I browse the web, I'd love to share it straight into gulliver so it can get indexed or return the data.

juliantoledo:
This is a great idea, we should definitely implement it.

andreban:
I think we can do this as part of rebuilding the PWA submission process. I've added an endpoint to allow us to do it asynchronously on #501 , and this could also be used to handle the share target.

andreban:
@fstanis, FYI

PR #507, opened 2018-07-16
Add Web Share Target API support

fstanis:
Relevant issue: #500

### DIFF
--- a/controllers/pwa.js
+++ b/controllers/pwa.js
@@ -16,6 +16,7 @@
 'use strict';
 
 const express = require('express');
+const dataFetcher = require('../lib/data-fetcher');
 const pwaLib = require('../lib/pwa');
 const libPwaIndex = require('../lib/pwa-index');
 const verifyIdToken = require('../lib/verify-id-token');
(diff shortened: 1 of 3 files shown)


controllers/pwa.js
```diff
@@ -152,18 +153,35 @@ router.get('/search', (req, res, next) => {
...
-    title: 'PWA Directory - Submit a PWA',
-    description: 'PWA Directory: Submit a Progressive Web Apps',
-    pwa: {},
-    action: 'Add',
-    backlink: true,
-    submit: true,
-    contentOnly: contentOnly
-  });
-  res.render('pwas/form.hbs', arg);
+router.get('/add', async (req, res, next) => {
```
andreban: We have recently changed to make the PWA submission using AJAX from the frontend (see https://github.com/GoogleChromeLabs/gulliver/pull/503). 

I think the manifestUrl parameter should be handled there as well. (and we should probably remove the leftovers of when it was done in the backend)

PaulKinlan:
Works for me! excited to see this land.

PR #516, opened 2018-09-03
Minor fixes

juliantoledo:


lib/color.js
```diff
@@ -56,7 +56,11 @@ function relativeLuminance(color) {
   let colorGreen;
   let colorBlue;
 
-  [colorRed, colorGreen, colorBlue] = parseColor(color.trim()).rgb;
+  try {
```
andreban: I think this code should throw an Error if the argument is invalid. The try/catch should be on the part of the code that is invoking this (the code invoking contrastRatio?)

juliantoledo: true, I will move it up to where this is used.

PR #517, opened 2018-09-12
Add share button

fstanis:
This PR adds a share button (using [`navigator.share`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share)) to PWA Directory.


public/js/ui/share-button.js
```diff
@@ -0,0 +1,62 @@
...
+
+  _getTitle() {
+    const pwaName = this._window.document.querySelector('#pwa-name');
+    if (!pwaName) {
+      return 'PWA Directory';
+    }
+    return pwaName.innerText.trim();
+  }
+
+  share() {
```
andreban: If I understood correctly, the share button is always available, but share fails if the browser doesn't support it. Would it be possible to simply hide the button if that's the case?

fstanis: It's disabled in `_init` if unavailable - I could hide it instead, but was worried about a "flash" of going from visible to invisible. Thoughts?

fstanis: Good point, done.

juliantoledo:
Very cool, do you have a screenshot of how it looks?

fstanis:
wxukaatwygg

The dialog itself is the platform-native one, e.g. on Android it will be the [share chooser](https://developer.android.com/training/sharing/send).

PR #526, opened 2018-12-03
Remove res.body.cancel

juliantoledo:
@andreban Can we just remove this? I cannot find a problem if I remove it.

### DIFF
--- a/lib/data-cache.js
+++ b/lib/data-cache.js
@@ -81,7 +81,6 @@ function set(key, value, lifetime) {
   return new Promise((resolve, reject) => {
     memcached.set(key, value, lifetime, err => {
       if (err) {
-        console.error(err);
         return reject(err);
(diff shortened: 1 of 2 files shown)


andreban:
This was part of the code when H2 support was added (https://github.com/GoogleChromeLabs/gulliver/commit/b04f6867bebd0f325cea2cb731fc3a440331b53a). I changed from res.text(), which would fully consume the content, to res.body.cancel() that actually cancels the stream.

Not sure why res.text() was added initially. Maybe to avoid leaving pending streams?

juliantoledo:
Yes, I most be because of the pending streams.

juliantoledo:
@andreban the spdy workaround is not needed anymore

andreban:
Awesome! thanks!

